PR #115, opened 2026-04-16
feat(plugin): add plugin system core framework with lifecycle management

PeterGuy326:
## Summary

- **What changed?**
  Added a complete plugin system for the CLI, including plugin manifest/loader, stdio transport, plugin updater (git-based install & update), plugin command registration, build lifecycle for binary distribution, persistent plugin config via `settings.json`, custom auth headers for third-party MCP servers, dev mode with skill sync, and default managed plugin bootstrapping. New packages: `internal/plugin`, `internal/transport`; extended `internal/app`, `internal/compat`, `pkg/config`. 25+ files changed, ~5500 lines added.

- **Why is this change needed?**
  Enables the CLI to be extended via third-party plugins (stdio/HTTP MCP servers), supporting install, update, scaffold, dev mode, and command overlay — making the tool ecosystem-ready without hardcoding every integration.

## Security Hardening (latest commit)

| Issue | File | Fix |
|-------|------|-----|
| Git URL protocol injection | `loader.go` | Reject `file://` and local paths, only allow https/ssh |
| ZIP symlink path traversal | `updater.go` | Reject symlink entries in ZIP extraction |
| Build output path traversal | `loader.go` | Validate `build.output` is relative, no `..` |
| Stdio command absolute path | `manifest.go` | Reject absolute paths in MCP server command |
| Env var injection | `loader.go` | Block dangerous env var names (PATH, LD_PRELOAD, etc.) |

## Verification

- [x] `make build`
- [x] `go vet ./...`
- [x] `make test` (unit tests all pass; integration/extensions skipped — plugin source not yet available)
- [x] `make policy`

## Notes

- Command surface has changed (new plugin subcommands) — `check-command-surface.sh --strict` has been run and passes.
- Plugin updater relies on `git clone` — requires git on the host machine.
- Default managed plugins list is currently empty (conference plugin source not yet published).
- Follow-up: plugin marketplace integration, plugin signing/verification, and plugin dependency resolution are not yet implemented.


PeterGuy326 (commented):
## Code Review — Plugin System Security Audit

### CRITICAL (must fix before merge)

#### 1. Command Injection in Hook Execution
**File:** `internal/plugin/hook_adapter.go:86`
```go
cmd := exec.CommandContext(timeoutCtx, "sh", "-c", h.entry.Command)
```
Hook commands from `hooks.json` are passed directly to shell. A malicious plugin can execute arbitrary commands.
**Fix:** Validate/whitelist hook commands, or parse into argv array instead of shell expansion.

#### 2. Command Injection in Build Command
**File:** `internal/plugin/loader.go:770`
```go
cmd = exec.Command("sh", "-c", build.Command)
```
Same issue — `build.command` from untrusted `plugin.json` goes straight to `sh -c`.
**Fix:** Parse command into argv, validate against known build tools, or sandbox execution.

#### 3. ZIP Extraction — Symlink Not Handled
**File:** `internal/plugin/updater.go:354-362`
Zip-slip protection exists but doesn't handle symlinks within ZIP files. A crafted ZIP could use symlinks to write outside the destination.
**Fix:** Reject symlink entries in ZIP, or resolve and validate symlink targets.

#### 4. Git URL Not Validated
**File:** `internal/plugin/loader.go:359`
```go
cmd := exec.Command("git", "clone", "--depth", "1", gitURL, cloneDir)
```
Accepts any URL including `file://` protocol (reads local files). No protocol whitelist.
**Fix:** Validate URL scheme is `https://` or `ssh://` only. Strip credentials before logging.

---

### HIGH (should fix soon)

#### 5. Environment Variable Injection
**File:** `internal/plugin/loader.go:615-632`
Plugin config values are injected as env vars without name validation. Could set `LD_PRELOAD`, `PATH`, etc.
**Fix:** Whitelist env var name patterns, block known dangerous names.

#### 6. Subprocess Leak on Abnormal Exit
**File:** `internal/transport/stdio.go`
StdioClient has no automatic cleanup. If CLI crashes, stdio subprocesses become orphans.
**Fix:** Use process groups, add defer-based cleanup, or signal-based shutdown hooks.

#### 7. Auth Tokens Stored in Plaintext
**File:** `internal/app/auth_registry.go`
Tokens stored as plaintext strings in memory. Risk of exposure via logs/errors.
**Fix:** Consider secure storage (keyring), sanitize tokens from error messages.

---

### MEDIUM

8. `copyDir` skips source symlinks but doesn't check destination for pre-existing symlinks
9. `saveSettings()` silently ignores write errors — user unaware of config loss
10. Manifest allows absolute paths in `command` and `build.output` fields
11. Updater mutex held during entire HTTP update loop — blocks concurrent CLI processes
12. Plugin name regex allows trailing/consecutive hyphens

---

### Build & Test Results

| Check | Result |
|-------|--------|
| `make build` | ✅ PASS |
| `go vet` | ✅ PASS |
| `make policy` | ✅ PASS |
| Unit tests (plugin/app/transport) | ✅ ALL PASS |
| Integration tests | ⚠️ Network timeout (not a code bug) |

**Summary:** Architecture is solid. Main concern is command execution security — critical for a system that installs third-party plugins.

coffeeBigSir (approved):

PR #245, opened 2026-05-09
feat(schema): add CLIAliases tool override + json_parse_strict transform

PeterGuy326:
## Summary

Two generic CLI envelope schema enhancements that close gaps surfaced by the `cli_to_mcp` test suite — without resorting to product-specific helpers in open core.

### 1. `CLIToolOverride.CLIAliases` (`registry.go` + `dynamic_commands.go`)

Lets a single MCP tool register additional cobra command aliases via envelope JSON. Plumbs through the existing `Route.Aliases → cobra.Command.Aliases` path; conflicts with siblings are silently skipped by cobra.

**Envelope usage**:
```json
"get_range": {
  "cliName": "read",
  "cliAliases": ["get"],
  "group": "range",
  ...
}
```
After this, both `dws sheet range read` and `dws sheet range get` resolve to the same `get_range` MCP tool.

### 2. `json_parse_strict` transform (`transform.go`)

Strict JSON variant of `json_parse` that does **not** fall back to YAML. Use when the upstream tool requires a structured array/object value and silently coercing malformed input to a scalar string would mask a real user error.

**Observed bug it fixes**: `dws sheet filter-view create --criteria 'NOT_VALID_JSON'` was being accepted by `json_parse` (YAML happily parsed the input as a scalar string), and the request reached the server with `criteria: "NOT_VALID_JSON"`, which the API silently dropped to `criteria: {}`. Result: the filter view was created with empty criteria and no error returned. With `json_parse_strict`, the CLI rejects malformed JSON before sending.

## Why these are schema enhancements, not product hardcode

- `CLIAliases` works for any product / any tool — sheet, wiki, aitable, chat, anyone wanting command-name aliases benefits.
- `json_parse_strict` is a generic transform option, not tied to filter-view.
- No product-specific code in `open-core`; the actual "sheet range get == sheet range read" declaration lives in the envelope JSON, not in Go.

## Validation against `cli_to_mcp` regression suite

Tested with `dws-wukong/auto-test/cli_to_mcp/testcases/run_all_tests.py --edition wukong --products wiki,sheet`:

| Round | Sheet pass rate | Wiki pass rate | Combined |
|---|---|---|---|
| Before envelope work (baseline) | 63.3% | 83.6% | 60.4% |
| After 5 rounds of envelope-only fixes | 93.4% | 83.6% | 92.0% |
| **After this PR** | **95.3%** | **85.5%** | **93.9%** |

The +1.9pp came from this PR: 4 `test_get_alias_*` tests, 1 `test_list_alias_ls`, and 1 `test_create_invalid_criteria_json` flipped from FAIL to PASS.

## Test plan

- [x] `go test ./internal/compat/...` (existing transform + dynamic_commands tests still pass)
- [x] Smoke test `dws sheet range get --node X --sheet-id Y` returns the same payload as `dws sheet range read --node X --sheet-id Y`
- [x] Smoke test `dws wiki member ls --space X` returns the same payload as `dws wiki member list --space X`
- [x] Smoke test `dws sheet filter-view create --criteria 'NOT_VALID_JSON' ...` now exits non-zero with a clear validation error (was previously silently accepted)
- [ ] Add unit tests for `CLIAliases` rendering to a `cobra.Command` (suggested follow-up if reviewers want stricter coverage)
- [ ] Add unit tests for `transformJSONParseStrict` (suggested follow-up)

## Backward compatibility

- `cliAliases` is `omitempty` — envelopes without it deserialize unchanged
- `json_parse_strict` is a new transform name — existing `json_parse` users are unaffected
- No breaking changes to existing CLI commands / behavior


PeterGuy326:
Superseded by #246. This PR was opened from the test/pre-mcp-discovery branch which carried unrelated pre-production endpoint redirects; #246 is a clean branch from main containing only the schema enhancements.

Issue #282, opened 2026-05-13
[server-side] MCP registry envelope schema strips mapsTo field — blocks --content-file / stdin shipping


PeterGuy326:
## Context

Follow-up to #277 / #278 (CLI-side `MapsTo` + `file_read` transform).

When publishing a doc envelope to the pre-production registry (`pre-mcp.dingtalk.com`) containing `mapsTo` per the new schema, the field is silently dropped server-side. Other new fields (`hidden`, `required`, `transform`, `mutuallyExclusive`, `requireOneOf`) all survive — only `mapsTo` is missing from what the client retrieves.

## Evidence

**Envelope posted to pre-production registry** (excerpt):

```json
"content": { "alias": "content", "mapsTo": "markdown" },
"content-file": { "alias": "content-file", "mapsTo": "markdown", "transform": "file_read" }
```

**What client cache actually receives** (`~/.dws/cache/.../market/servers.json` after `dws-pre cache refresh`):

```json
"content": { "alias": "content" },
"content-file": { "alias": "content-file", "transform": "file_read" }
```

`mapsTo` is gone on both flags. Verified with `dws-pre schema "doc update"` and the local cache dump — both confirm the same omission.

## Impact

`mapsTo` is the mechanism that lets multiple sibling CLI flags feed one MCP parameter (e.g. `--content` literal + `--content-file` path-with-`file_read` both landing on the upstream `markdown` parameter). Without server-side support:

- The CLI registers `--content` / `--content-file` as if they were standalone MCP parameters
- Values land on `params[\"content\"]` / `params[\"content-file\"]` instead of `params[\"markdown\"]`
- MCP service rejects (markdown required, gets nothing)
- Result: `dws doc update --content-file ./x.md` is unusable end-to-end despite PR #278 shipping correct client behaviour

## Reproduction

```bash
# 1. Post envelope with mapsTo to pre-production
# 2. Refresh client cache
~/.local/bin/dws-pre cache refresh

# 3. Read back the envelope from cache
python3 -c \"
import json
c = json.load(open('$HOME/.dws/cache/default_default/market/servers.json'))
for s in c['servers']:
    cli = s.get('cli')
    if cli and cli.get('id') == 'doc':
        flags = cli['toolOverrides']['update_document']['flags']
        for fn, fd in flags.items():
            print(f'{fn}: keys={list(fd.keys())}')
        break
\"

# Expected output if mapsTo were preserved:
[...]

PR #278, opened 2026-05-13
feat(transform): add file_read transform for UTF-8 file / stdin content

PeterGuy326:
## Summary
Adds a new `file_read` transform to `compat.ApplyTransform`. Refs #277 (UX gap). Scope reduced — see "Scope decision" below; `MapsTo` deferred to a follow-up PR.

## What this PR ships

| File | Change | Why |
|---|---|---|
| `internal/compat/transform.go` | New `file_read` transform: reads UTF-8 file at the flag's value, or stdin when value is `-`. Errors surface as validation errors (exit 2). | Foundational primitive for any future flag that carries a path / stdin pointer to string-typed MCP content. Dormant until referenced by an envelope. |
| `internal/compat/transform_test.go` | 6 new tests | literal file / stdin via dash / missing file / non-UTF-8 / empty path / non-string input |

## Scope decision — `MapsTo` field intentionally NOT included

The original proposal in #277 paired `file_read` with a new `CLIFlagOverride.MapsTo` envelope field, so multiple sibling CLI flags (e.g. `--content` literal + `--content-file` path-with-file_read) could both feed one MCP parameter (e.g. `markdown`).

`MapsTo` was descoped from this PR to keep the change surface minimal — `file_read` is a self-contained transform that has value on its own, and shipping the routing field is a separate schema concern that needs its own design + tests + envelope round-trip validation.

The `MapsTo` field is tracked separately in #282 (re-scoped from "server strips mapsTo" to "client struct missing field + server round-trip needs validation" — see #282 update comment for details). Once that lands, a follow-up PR can wire `--content-file` end-to-end using this `file_read` transform.

The `--content` half of #277's user value is already shipping via an **`aliases`-based envelope** (uses only pre-existing schema fields, no client change required, validated green in pre-prod).

## Test plan

- [x] `go test ./internal/compat/... ./internal/market/...` — green
- [x] `go vet ./internal/compat/... ./internal/market/...` — clean
- [x] `go build ./...` — clean
- [x] Backward-compat: net diff is +1 case in `ApplyTransform` + 1 new helper + tests; no behaviour change for existing transforms or envelopes

## Cross-references

- #277 — original UX gap (Wukong has `--content / --content-file / stdin`, open does not)
- #282 — `MapsTo` follow-up (client struct field + dispatch routing + server round-trip validation); when resolved, a follow-up PR re-introduces `MapsTo` and the doc envelope upgrades to use both `MapsTo` + this `file_read` to expose `--content-file`

Issue #277, opened 2026-05-13
[doc][UX] dws doc update should align with wukong: support --content / --content-file / stdin


PeterGuy326:
## Problem

Open-source `dws doc update` only exposes a single `--markdown` string flag, which is awkward for non-trivial content:

```bash
# Today (open-source v1.0.26)
dws doc update --node DOC_ID --markdown "# header
- bullet1
- bullet2"   # multi-line via shell escape; tables get ugly fast
```

The Wukong edition (v0.2.67) has already migrated to a richer surface backed by the **same** MCP tool (`doc.update_document`):

```bash
dws doc update --node DOC_ID --content "# 短文本"          # short literal
dws doc update --node DOC_ID --content -                   # from stdin
dws doc update --node DOC_ID --content-file ./part.md      # from file (UTF-8, recommended for long/multi-line/tables)
dws doc update --node DOC_ID --content-file ./a.md --mode append
```

`dws schema "doc update"` confirms the upstream MCP parameter is still `markdown` — this is purely a CLI ergonomic layer, no MCP schema change needed.

## Why it matters

1. **AI Agent friendliness**: `--content-file` + `--mode append` is the only safe path for >2KB content (shell-escape limits, table formatting, special chars). All Agent recipes in the skill docs assume it works on both editions.
2. **Cross-edition parity**: Open / Wukong now diverge on a flagship product. Users switching editions hit silent breakage.
3. **Surface stability**: adding `--content` / `--content-file` as new flags is non-breaking; `--markdown` can be kept as a hidden alias.

## Proposal

Two-step rollout:

**Step 1 — this issue (CLI ergonomic layer, this repo):**
- Extend `compat.ApplyTransform` with a `file_read` transform (read UTF-8 file at the flag's value, return its contents; `-` means stdin)
- Extend `market.CLIFlagOverride` with a `MapsTo` field, so multiple CLI flags can map to one MCP parameter (e.g. `content` + `content-file` both `mapsTo: "markdown"`)
- Backward-compat: keep `--markdown` registered as a hidden alias

**Step 2 — follow-up (server envelope, outside this repo):**
- Update the `doc` MCP server's overlay to declare `content` / `content-file` aliases via the new mechanism

## Acceptance

- `dws doc update --help` shows `--content`, `--content-file`, `--mode`, `--node`
- `dws doc update --markdown ...` still works (hidden), prints a deprecation hint in --help epilog
[...]

PR #288, opened 2026-05-14
feat(schema): add CLIFlagOverride.MapsTo for sibling-flag routing (#277 #282)

PeterGuy326:
## Summary

Wires `CLIFlagOverride.MapsTo` so a CLI flag's final value (post-transform or literal) can be routed into a **different** MCP parameter slot than its own property name. Unlocks #277's `--content-file` end-to-end pattern without breaking existing flags. Pairs with the `file_read` transform from #278.

## Why

Today `buildOverrideBindings` writes `params[propertyName] = value` (or `transformed_value`). One MCP parameter can only be sourced from one CLI flag. To expose `--content-file ./a.md` (file_read transform) alongside the existing literal `--content "# 标题"`, both feeding the upstream `markdown` parameter, we need a per-flag routing field.

`MapsTo` is that field. It was previously misdiagnosed as "server strips it" in #282 — root cause re-investigation (see #282 update comment) confirmed the client struct simply never declared the field, so `json.Unmarshal` was dropping it.

## What ships

| File | Change |
|---|---|
| `internal/market/registry.go` | Add `MapsTo string \`json:"mapsTo,omitempty"\`` on `CLIFlagOverride` with a doc-comment explaining the routing semantic and the pairing with tool-level `MutuallyExclusive` |
| `internal/compat/dynamic_commands.go` | Extend `transformEntry` with `mapsTo`; add `mapsToRoutes` for routing-only flags (no transform); update the normalizer closure to write to `params[MapsTo]` and `delete` the source key when MapsTo is set. `enum_map` `_default` injection respects MapsTo too. |
| `internal/compat/transform.go` | Rewrite the `transformFileRead` doc-comment from "Use with MapsTo" placeholder to "Typical envelope use is paired with CLIFlagOverride.MapsTo" — now describing live behaviour |
| `internal/compat/dynamic_commands_test.go` | 4 new tests (see Test plan) |

## What is **not** in scope (intentionally)

- **`MutuallyExclusive` per-flag** — already exists at tool level via `CLIToolOverride.MutuallyExclusive [][]string` + cobra `MarkFlagsMutuallyExclusive`. No duplicate machinery added; envelope authors should keep using the tool-level form for guarding "set one, not both" sibling flags. The new test `_BothSetIsRejectedByCobra` documents this composition.
- **`RequireOneOf`** — same story, already at tool level.
- **Envelope changes** — out of scope for the client repo; the pre-prod doc envelope (currently using an `aliases`-only workaround per #282) needs a separate update to actually route `--content` / `--content-file` via this field.

## Test plan

- [x] `go test ./internal/compat/... ./internal/market/...` — green
- [x] `go vet ./...` — clean
- [x] `go build ./...` — clean
- [x] New tests: `TestBuildDynamicCommands_MapsTo_*`
  - `WithoutTransform` — literal --content → params[markdown], source key deleted
  - `WithFileReadTransform` — --content-file path → params[markdown] (file contents), source key deleted
  - `SiblingFlagsExclusiveSetOne` — two sibling flags both mapsTo same target, only one set: clean routing
  - `BothSetIsRejectedByCobra` — both siblings set: rejected by tool-level MutuallyExclusive (regression check on composition)
- [x] Backward-compat: empty `MapsTo` preserves `params[propertyName] = value` write semantics for every existing envelope. All pre-existing dynamic_commands tests pass unchanged.

## Pre-prod end-to-end (post-merge)

Once merged, the doc envelope on `pre-mcp.dingtalk.com` needs a separate push to actually use `mapsTo`. After that:

```bash
~/.local/bin/dws-pre cache refresh
~/.local/bin/dws-pre doc update --node DOC_ID --content-file ./long.md --mode append
# expected: upstream MCP receives params[markdown]=<file contents>
```

If `mapsTo` survives the round-trip in the raw HTTP response, #282 closes. If it gets stripped, #282 reopens with concrete evidence and a server-side ask.

## Cross-references

- #277 — original UX gap (Wukong has `--content` / `--content-file` / stdin; open does not). This PR closes Step 1b.
- #278 (merged) — `file_read` transform; this PR is its routing companion.
- #282 — root cause re-investigated; this PR is the client-side fix. Server-side round-trip validation pending post-merge.

### DIFF
--- a/internal/compat/dynamic_commands.go
+++ b/internal/compat/dynamic_commands.go
@@ -545,10 +545,19 @@ func buildOverrideBindings(override market.CLIToolOverride) ([]FlagBinding, Norm
 	var bindings []FlagBinding
 	type transformEntry struct {
 		paramName     string
+		mapsTo        string
 		transform     string
 		transformArgs map[string]any
 	}
 	var transforms []transformEntry
+	// mapsToRoutes captures flags that only need value-routing (no transform)
+	// — e.g. a literal --content flag that mapsTo "markdown". The dispatch
+	// loop moves params[paramName] → params[mapsTo] after CLI binding.
+	type mapsToRoute struct {
+		paramName string
+		mapsTo    string
+	}
+	var mapsToRoutes []mapsToRoute
 	type envDefaultEntry struct {
 		paramName string
 		envVar    string
@@ -677,8 +686,18 @@ func buildOverrideBindings(override market.CLIToolOverride) ([]FlagBinding, Norm
 		if flagOverride.Transform != "" {
 			transforms = append(transforms, transformEntry{
 				paramName:     paramName,
+				mapsTo:        strings.TrimSpace(flagOverride.MapsTo),
 				transform:     flagOverride.Transform,
 				transformArgs: flagOverride.TransformArgs,
+			})
+		} else if mt := strings.TrimSpace(flagOverride.MapsTo); mt != "" {
+			// mapsTo without transform: just route the literal value into a
+			// different MCP parameter slot. Common case is --content (literal
+			// string) mapping to MCP parameter markdown, alongside a sibling
+			// --content-file (transform: file_read) mapping to the same slot.
+			mapsToRoutes = append(mapsToRoutes, mapsToRoute{
+				paramName: paramName,
+				mapsTo:    mt,
 			})
 		}
 		if flagOverride.EnvDefault != "" {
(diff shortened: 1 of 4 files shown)


PeterGuy326:
Superseded by #291.

本 PR 的 MapsTo 客户端实现已经 cherry-pick 到 #291（基于 main，不带 test 分支的 staging hack）。预发端到端验证 6 条用例全部通过，证据见 #291 PR body 的 Validation 表：
- mapsTo 字段从预发原始 HTTP 响应里正确 round-trip（30 次出现）
- --content / --content-file / stdin 三种入口都精确路由到 params.markdown
- 互斥校验 + requireOneOf 两道护栏都生效

关闭此 PR，head 分支 `feat/maps-to-routing` 保留不删（方便回溯）。

PR #332, opened 2026-05-20
feat(skill): dws skill setup with mono/multi modes

PeterGuy326:
## Summary

Introduces `dws skill setup` command with **two installation modes**:

- **mono** (default, recommended) — single `dws` skill covering all DingTalk products. Backward compatible with current users.
- **multi** (experimental) — 29 product-scoped skills (`dingtalk-aitable` / `dingtalk-calendar` / ...) installed as sibling Agent skills.

Switching is bidirectional with automatic mutual-exclusion cleanup. Zero impact on existing `curl ... | sh` users — they keep getting exactly the same `~/.claude/skills/dws/SKILL.md` they have today.

> **This was originally split into 4 separate PRs (#328–#331). After review feedback that the cross-fork stacked PRs were producing overlapping diffs and adding review overhead without independent value, they were superseded and consolidated into this single PR. The 4 commits are preserved — please review commit by commit.**

### Commit 1: `feat(skill): reorganize skills/ into mono subdir and add dws skill setup command`
Foundation. No behavior change for existing users.
- `skills/* → skills/mono/*` (git rename, history preserved)
- New `dws skill setup` cobra subcommand with [huh](https://github.com/charmbracelet/huh) interactive prompts
- `install.sh` / `install.ps1` / `post-goreleaser.sh` point at `skills/mono/`
- `--mode multi` returns "not yet enabled" until commit 3
- 47 files / +87/-7

### Commit 2: `feat(skills): port multi-skill layout into skills/multi/ (29 product skills)`
Pure content port from `dws-wukong` (`feature/multiSkill` branch). Zero code.
- 29 dingtalk-* directories under `skills/multi/`
- Each contains `SKILL.md` + `references/` + (where applicable) `scripts/`
- Verified: 120 files / no `dws-wukong` path references / no absolute paths
- 120 files / +26K lines

### Commit 3: `feat(skill): wire up multi mode + install.sh TTY interaction + README docs`
Connects mono+multi into a usable feature.
- `dws skill setup --mode multi` enabled
- 6 new helper functions including bidirectional mutual-exclusion cleanup (installing mono cleans `dingtalk-*` siblings; installing multi cleans `~/.claude/skills/dws/`)
- `DWS_SKILL_MODE` env var + TTY interactive prompt in `install.sh` / `install.ps1`
- `curl ... | sh` keeps falling back to mono silently (no TTY → default mono)
- README.md / README_zh.md gain mono vs. multi comparison + 4 invocation paths
[...]

PR #337, opened 2026-05-21
test: static skill-command dispatch verifier (Layer A)

PeterGuy326:
## Summary

Adds a Go test (`test/skill_static/skill_static_test.go`) that statically verifies every backtick `dws ...` command referenced in `skills/**/*.md` actually dispatches as a cobra sub-command in open-source dws v1.0.30. This is **Layer A** of the two-layer skill verification matrix being introduced for the mono/multi skill mode PR (#332).

## Why

PR #332's manual audits caught many real-but-not-runnable commands in skill docs (e.g. `dws sheet copy_sheet`, `dws minutes detail`, `dws chat message recall`). With 426 extracted commands across 20 products, manual audits don't scale and regress on every PR.

This test makes the regression-protection automatic:
- Any PR that adds a doc command pointing at a non-existent cobra sub-path will fail this test.
- Any PR that removes a real cobra command without updating skill docs will surface the broken references on next run.

## How the verifier works

1. Recursively scan `skills/mono/SKILL.md`, `skills/mono/references/products/*.md`, `skills/multi/*/SKILL.md`, and `skills/multi/*/references/*.md` for backtick `dws ...` patterns.
2. Skip noise — markdown optional flags (`[--foo]`), templated placeholders containing spaces (`<某个值>`), shell pipes / command substitution / `&` background ops, `>`/`...`/etc.
3. Parse the sub-path (tokens before the first `--flag`), preserving double-quoted strings (`--query "OAuth2 接入"`) as opaque tokens.
4. Run `dws <sub-path> --help` and assert either:
   - The Usage line literally starts with the sub-path, OR
   - The parent command's `Available Commands:` block lists the leaf.
5. Small allowlist for intentional documentation features:
   - `dws calendar list` — anti-pattern lines warning "CLI doesn't have this; don't invent it"
   - `dws minutes detail/info/summary/transcription` — appears in the 「高频错误命令 vs 正确命令」 comparison table
   - `dws report inbox` — explicit "禁止编造" warnings
   - `dws skill add / dws skill find` — backward-compat stubs (cobra omits from Available Commands but they're real)

## Cleanup driven by bringing the test green

Two stale alias claims surfaced and were removed:
- `skills/mono/references/products/chat.md` line 3 — claimed `dws im` is an alias for `dws chat`; doesn't exist in open-source dws.
- `skills/multi/dingtalk-report/references/report.md` line 3 — same for `dws log`. Replaced with a clarifier about the OA weekly-report semantic.

## Run

```
PATH=/path/to/dws-binary:$PATH go test ./test/skill_static/... -v
```

Output:
```
=== RUN   TestSkillCommandsDispatch
    Extracted 426 command references from skill docs
    ✓ All extracted commands dispatch cleanly (sub-paths registered, anti-patterns whitelisted)
--- PASS: TestSkillCommandsDispatch (37.23s)
```

Skips gracefully when `dws` is not on PATH so unrelated `go test ./...` runs aren't blocked.

## Future work (out of scope for this PR)

- Wire this into the existing CI Test job (it already runs Go tests; just needs PATH setup so the freshly-built `dws` is discoverable).
- Add a parallel test asserting mono ⇄ multi command set parity.
- Layer B (real-tenant golden-case integration tests) will live in `dws-wukong/test/skill-integration` since it needs internal test data and a fixture-prepared tenant.

## Test plan

- [x] `go test ./test/skill_static/... -v` passes on `feature/skill-setup` HEAD (`23fb804`) after this branch is rebased on top
- [x] Adding `dws fake command` to any skill doc → test fails with a clear "not in Available Commands of dws" report
- [x] Skip path: removing `dws` from PATH → test skips with a clear message, no false fail

Net: 4 files / +335 / -5.


PeterGuy326:
**Update**: end-to-end live verifier (Layer B) caught one additional real fail that the Layer A static check missed.

While dogfooding the verifier infrastructure I built a parallel **live runner** that actually executes every backtick `dws ...` command (with placeholders substituted to safe live/dummy values, and `--dry-run` auto-injected for writes), then classifies the response. The runner deliberately includes anti-pattern allowlists for documentation `错例 vs 正解` tables and `[禁止]` blocks.

**Result on PR #332's tree**: 1 real failure remained.

```
❌ skills/multi/dingtalk-calendar/references/calendar.md:88
   dws calendar event list --calendar-id primary
   → unknown flag: --calendar-id
```

`dws calendar event list` / `event get` in open-source dws v1.0.30 do **not** expose `--calendar-id`. The flag is a wukong-internal extension. Without this fix any agent reading the calendar skill would compose `--calendar-id primary` and hit immediate cobra rejection.

Commit 890276b removes all `--calendar-id` references (4 example lines + 2 cross-reference notes + flag declarations) and clarifies inline that the OSS path defaults to primary calendar.

**Final live-runner state**: 0 `unknown flag` / 0 `unknown command` across 297 unique commands extracted from the skill tree.

(Live runner script and detailed run script are kept in /tmp for now — once this PR lands and the Layer A Go test is wired into CI, I'll fold the live runner into a dws-wukong companion since it needs the real test tenant.)

PeterGuy326:
**Update**: Layer B end-to-end test added (commit 712edfe) — fully open-source, no dws-wukong dependency.

## Two-layer matrix, entirely in this repo

| Layer | What it verifies | Mechanism | Tenant needed |
|-------|------------------|-----------|---------------|
| **A static** `test/skill_static` | every skill doc command exists in cobra | parse `dws X --help` | No |
| **B mock e2e** `test/skill_e2e` | every command dispatches cleanly through cobra → envelope → dispatcher | real `dws cmd --mock` invocation | No (`--mock` short-circuits before network) |

## Layer B mechanics

The dws binary's built-in `--mock` flag returns a deterministic shape for every command that resolves cleanly:

```json
{ "_mock": true, "_tool": "<canonical>", "result": [], "success": true }
```

So Layer B catches **everything Layer A catches plus**:
- envelope tool-mapping failures (helper override drift)
- flag-overlay / alias mapping bugs
- write-vs-read auto-routing

…without needing real tenant data. `--mock` short-circuits before the network call.

## Result

```
=== RUN   TestSkillCommandsMockDispatch
    Extracted 312 refs, 218 unique runnable commands
    ✓ All 218 unique skill commands dispatch cleanly under --mock
--- PASS: TestSkillCommandsMockDispatch (57.53s)
```

Both layers skip gracefully when `dws` is not on PATH, so unrelated `go test ./...` runs aren't blocked.

## What does NOT live in this repo

- Real-tenant 'golden case' integration tests — would leak credentials or be flaky on tenant drift.
- Live-tenant verifier scripts — maintainers can keep these locally if they want to dogfood with a real account.

PR is now ready to merge: it ships **both** layers of skill-command verification with **zero external dependency**.

PeterGuy326:
Superseded by #332 — moved both test layers (skill_static + skill_e2e) and the calendar --calendar-id fix into PR #332's branch so they ride along with the skill changes they verify. Closing here to avoid duplicate review surface.

PR #416, opened 2026-06-05
fix(upgrade): honor --dry-run as preview-only instead of silently upgrading

### LINKED ISSUE #364: bug: dws upgrade --dry-run performs the actual upgrade
## Summary

`dws upgrade --dry-run --version v1.0.32 --yes` appeared to perform the actual upgrade instead of only previewing the operation.

## Environment

- macOS arm64
- Starting version: `dws version v1.0.30 (754b0df, 2026-05-19T09:47:47Z)`
- Target/latest version: `v1.0.32`

## Command run

```bash
dws upgrade --dry-run --version v1.0.32 --yes --format json
```

## Expected behavior

Because `--dry-run` is documented as preview-only, I expected no local files/binaries/skills to be changed.

## Actual behavior

The command output showed real upgrade steps:

```text
[1/5] 备份当前版本... ✓
[2/5] 下载 dws-darwin-arm64.tar.gz ✓
[3/5] 校验 dws-darwin-arm64.tar.gz... ✓
[4/5] 解压并验证... ✓
[5/5] 替换并安装... ✓
✔ 升级完成  v1.0.30 → v1.0.32
```

Afterward:

```bash
dws --version
# dws version v1.0.32 (86ff7e2, 2026-05-25T11:59:11Z)
```

A rollback backup was also created at the normal backup location.

## Extra observation

After this binary-style upgrade, `npm list -g --depth=0 dingtalk-workspace-cli` still reports `dingtalk-workspace-cli@1.0.30`, while `dws --version` reports `v1.0.32`. That may be expected if the upgrader replaces the vendored binary inside the npm-installed wrapper, but it can be confusing during diagnostics.

## Request

Please make `--dry-run` non-mutating for `dws upgrade`, or disable/reject that flag for upgrade commands if true dry-run is not supported. It would also help if diagnostics clarified the difference between npm package metadata and the embedded `dws` binary version.

jackjin1997:
## Summary

- **What changed?** `dws upgrade` now honors `--dry-run`: it resolves the target release and platform asset (so "already latest" / a missing build for the platform is still reported), prints the 1–5 steps it *would* perform, and returns **before any side effect** — no backup, no download, no replace. `--dry-run` is also advertised in the command examples.
- **Why?** `newUpgradeCommand` registered no `--dry-run` flag and never read the global persistent one, so `--dry-run` fell through to `runUpgrade` and performed a **real, irreversible upgrade** (download + binary replace). This directly contradicts the flag's documented contract (`预览操作内容，不实际执行`) and is dangerous: users running a "preview" actually got upgraded.

Fixes #364

## Verification

- [x] `make build`
- [ ] `make lint` — `go vet` passes; `staticcheck` not installed in my local env (CI covers it)
- [x] `go test ./internal/app/` (incl. new `TestWriteDryRunPlan_*`, updated help test)
- [x] `make policy`

## Notes

- **Semantics chosen:** pure preview (zero side effects), aligning with the flag's own description. The dry-run still performs the read-only version/asset resolution so it surfaces "no matching asset for this platform" and "already latest". The alternative (download+verify, skip only the replace) was considered but rejected because it would still write to disk, contradicting "不实际执行".
- The preview renderer (`writeDryRunPlan`) is writer-injected and side-effect-free for testability.
- **Scope:** limited to the reported `dws upgrade` path. `--dry-run` + `--rollback` is the same "mutating op ignores dry-run" class and could be a small follow-up, but is intentionally out of scope here to keep the change atomic.

### DIFF
--- a/internal/app/upgrade.go
+++ b/internal/app/upgrade.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -56,6 +57,7 @@ func newUpgradeCommand() *cobra.Command {
   dws upgrade --list --all       # 列出所有版本
   dws upgrade --version v1.0.5   # 升级到指定版本
   dws upgrade --rollback         # 回滚到上一版本
+  dws upgrade --dry-run          # 仅预览升级步骤，不实际执行
   dws upgrade -y                 # 跳过确认直接升级`,
 		Args: cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
@@ -68,6 +70,7 @@ func newUpgradeCommand() *cobra.Command {
 			}
 
 			yes, _ := cmd.Flags().GetBool("yes")
+			dryRun, _ := cmd.Flags().GetBool("dry-run")
 			format := resolveUpgradeFormat(cmd)
 
 			if flagList {
@@ -88,6 +91,7 @@ func newUpgradeCommand() *cobra.Command {
 				force:         flagForce,
 				skipSkills:    flagSkipSkills,
 				yes:           yes,
+				dryRun:        dryRun,
 			})
 		},
 	}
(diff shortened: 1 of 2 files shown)


PeterGuy326 (approved):

PR #447, opened 2026-06-10
fix(cli): degrade to built-in commands when the dynamic command build panics

PeterGuy326:
## Problem

The dynamic command tree is built from cached discovery data **before** Cobra dispatches any command (`NewRootCommandWithEngine` → `newLegacyPublicCommands` → `loadDynamicCommands`, ahead of `root.ExecuteC()`). A panic during that build therefore aborts **every** invocation — including `dws cache refresh` and `dws --version`.

This is exactly the pre-1.0.32 lock-out: a poisoned cache triggered `chat_permission_grant flag redefined: params`, and the only recovery was telling users to manually run `rm -rf ~/.dws/cache/default_default/tools/*`. The duplicate-flag class itself has since been fixed at the builder level (reserved-name skip + `Lookup` dedup in `ApplyBindings` / the Detail-schema flag path), but the structural failure mode remained: any *future* panic class in the cache-driven build would brick the CLI again, with only the top-level `recover()` in `Execute()` printing `internal panic` and exiting.

## Fix

Wrap the envelope-driven build in a local `recover()` (`buildEnvelopeCommandsSafe` in `internal/app/legacy.go`). On panic the CLI now:

1. logs the panic via `slog.Error`,
2. prints a stderr hint:
   ```
   Warning: building product commands from the local discovery cache failed: <panic>
   Product commands are temporarily unavailable; utility commands still work.
   Run 'dws cache refresh' to rebuild the cache.
   ```
3. falls back to the hardcoded helper commands.

Worst case is now a degraded session with product commands missing, while `auth` / `cache` / `doctor` / `version` / `upgrade` and the helper commands stay alive — so `dws cache refresh` can rebuild the poisoned cache and the user (or an agent) self-heals without deleting files by hand.

## Tests

- `TestNewLegacyPublicCommandsPanicFallsBackToHelpers`: simulates a build panic via a test seam, asserts no panic propagates, helper commands are returned, and the stderr hint mentions `dws cache refresh`.
- `TestNewLegacyPublicCommandsNoPanicKeepsDynamicPath`: asserts the guard is transparent on the happy path.
- Full `go test ./internal/app/` passes; smoke-tested the built binary (`dws version` with a fresh cache dir).

### DIFF
--- a/internal/app/legacy.go
+++ b/internal/app/legacy.go
@@ -16,6 +16,7 @@ package app
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"log/slog"
 	"net"
 	"net/http"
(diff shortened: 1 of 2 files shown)


PeterGuy326:
**Verified end-to-end with real cache data — no code instrumentation.**

Found a live panic vector that still exists on `main`: the envelope's `shorthand` field is passed straight to pflag's `StringP`, and pflag panics on duplicate (or multi-char) shorthands with no guard on that path. So a registry snapshot where two flags of one tool share `"shorthand": "x"` reproduces the exact pre-1.0.32 lockout class with a byte-correct cache file (written via `cache.Store.SaveRegistry`, pointed at with `DWS_CACHE_DIR`).

**Baseline (`main` @ eaa60f9) — total lockout reproduced, every command dies:**

```
$ dws version
Error: internal panic: unable to redefine 'x' shorthand in "boom" flagset: it's already used for "alpha" flag
exit=5
$ dws cache refresh      # the repair command itself is dead
Error: internal panic: unable to redefine 'x' shorthand in "boom" flagset: ...
exit=5
```

**This branch (882a8d4) — same poisoned cache, degraded mode + self-heal:**

```
$ dws version
Warning: building product commands from the local discovery cache failed: unable to redefine 'x' shorthand ...
Product commands are temporarily unavailable; utility commands still work.
Run 'dws cache refresh' to rebuild the cache.
Version:        dev
exit=0
$ dws cache refresh
[OK] 缓存刷新完成：已刷新 22 个服务，失败 0 个
exit=0
$ dws version            # fully recovered, no warning
Version:        dev
exit=0
$ dws doc --help         # product commands are back
钉钉文档（搜索 / 浏览 / 读写 / 上传下载 ...）
```

The self-heal loop closes: panic → degraded session with the hint → `cache refresh` overwrites the poisoned snapshot → next invocation is fully functional. No manual `rm -rf ~/.dws/cache/...` needed.

Side note: the duplicate-shorthand registration found while building this repro is a real unguarded path in `internal/compat/registry.go` (`register(primary, true, ...)`) — worth a follow-up dedup guard like the long-flag one, independent of this PR.

audanye-sudo (approved):

PR #514, opened 2026-06-26
fix(report): native --contents-file / stdin for entry submit + skill sync

PeterGuy326:
## Problem

`dws report entry submit --contents-file <f>` (and `--contents -` stdin) silently submitted `contents: [null]` and the report failed. Only inline `--contents '[...]'` worked — which is why the inline-only helper `report create` succeeded while `report entry submit --contents-file` did not.

Root cause is an architecture mismatch, not a single bad transform:

- `report entry submit` is an **envelope-driven** command (MCP tool `create_report`). Its `--contents` flag has `json_parse` (required); the sibling `--contents-file` has `omitWhen empty` but **no transform / mapsTo**, so its value mapped to the unused `contentsFile` param while the real `contents` stayed empty.
- The wukong reference reads the file/stdin **natively** in Go (`resolveReportContentsFromFlags`, priority `--contents-file` > `--contents -` > inline). The open edition already mirrors wukong on the read side via post-merge decorator hooks (`AttachReportListReadableEnrichment`), so this follows the same idiom.

## Fix (no discovery-config change)

`internal/compat/report_hooks.go` — a build-time hook on the `create_report` leaf, wired beside the existing mail/todo/calendar/attendance hooks:

1. Clears the individual `required` on `--contents` and declares a `contents` / `contents-file` **one-of** group, so a `--contents-file`-only invocation is no longer rejected at parse time (validated before PreRunE).
2. A chained PreRunE reads the file/stdin (10MB cap, UTF-8 check) and writes the JSON back into `--contents` so the downstream `json_parse` decodes it uniformly; `--contents-file` is then cleared.

The existing cedar `contentsFile` flag declaration is **kept** — it is what registers the `--contents-file` flag the hook consumes. No Diamond / discovery change is required.

## Skill sync

The `multi/dingtalk-report` skill tree was two versions behind `mono` (taught deprecated flat aliases; falsely claimed `report inbox` was unimplemented). Re-aligned SKILL.md + 05-reporting.md + mono intent-guide.md to the canonical `template get` / `entry submit --contents-file` / `inbox list` / `outbox list` / `entry get` / `entry stats`.

## Verification

End-to-end against the **live cedar envelope** via `--dry-run`:

- `--contents-file <f>` (no `--contents`) → no required error; `contents` populated with the parsed array.
- `--contents -` (stdin) → populated from stdin.
- inline `--contents` → left untouched.
- neither → `at least one of the flags in the group [contents contents-file] is required`.

11 new unit tests in `report_hooks_test.go`; `go test ./internal/compat/ ./internal/helpers/ ./internal/app/` green. Full wukong01 sandbox eval in progress.


PeterGuy326:
## Verification

### 1. Unit + dry-run (live cedar envelope)

`go test ./internal/compat/ ./internal/helpers/ ./internal/app/` green; 11 new tests in `report_hooks_test.go`.

End-to-end `report entry submit` against the live cedar envelope via `--dry-run`:

| Input | Result |
|---|---|
| `--contents-file <f>` (no `--contents`) | no required error; `contents` populated with the parsed array |
| `--contents -` (stdin) | populated from stdin |
| inline `--contents '[...]'` | left untouched |
| neither | `at least one of the flags in the group [contents contents-file] is required` |

Before the fix the same `--contents-file` invocation sent `contents: [null]` and the report failed.

### 2. Full cli_to_mcp eval (wukong01 sandbox, PR binary)

Built this branch and ran the full open-edition suite as wukong01 (PAT scopes fully granted first):

- **report: 4 passed / 0 failed → 100%** — the fixed product, zero regression.
- 100% products: report, calendar, attendance, oa, doc, minutes, aisearch, drive, live.
- Aggregate 2358 cases, 77.6% — dragged down almost entirely by 363 `sheet` TIMEOUT errors (known infra issue, needs direct pytest) and the mail org-restriction (`暂不支持阿里巴巴邮箱`), neither related to this change.
- Every non-100% product was triaged to a pre-existing cause (sheet/aitable server timeout, contact HR-roster account permission, todo sub-task tool not yet exposed, chat group-history permission, wiki proxy route, devdoc pre-prod stub) — **none caused by this change**, which only adds a report-scoped compat hook plus one no-op-for-non-report hook call.

### 3. Discovery

No discovery-config change required. The cedar `contentsFile` flag declaration is kept — it is what registers the `--contents-file` flag the hook consumes; the hook resolves its value natively before `json_parse` runs.

PeterGuy326:
## Full-command audit follow-up (@file / @- generalization)

After the report fix, swept the whole command tree (643 nodes / 523 leaves) and all structured JSON flags for the same bug class.

**Healthy:** command tree has zero panics / flag-redefine / nil-pointer; the only non-zero exits are intentional deprecation-redirect stubs. The `*-file` content-reader bug existed only on report (fixed); `doc --content-file` is the correct reference; other `*-file` flags are metadata (name/size/id/type).

**Two findings, both fixed in f0ca47ec:**

1. **Misleading message (bug):** `json_parse`'s error advertised "pass `@path/to/file.json`" but no `@file` resolution existed.
2. **No file escape-hatch (gap):** structured JSON flags accepted inline only — long payloads (many records, big 2D ranges, filter criteria) hit shell-quoting hell.

Fix: `@file` / `@-` becomes a uniform convention for structured flags.
- `json_parse` / `json_parse_strict` expand a leading `@` before parsing (`@-` = stdin, `@<path>` = file). Covers every envelope-driven structured flag and makes the hint truthful.
- `sheet`'s shared `sheetParseJSONFlag` routes through `cli.ResolveInputSource`, so `--values` / `--criteria` / `--sort-keys` get the same support (sheet helpers shadow the envelope, so the transform fix alone wouldn't reach them).

Not changed: aitable already has `--records-file`; doc already has `--content-file` (its `--element` JSONML pipeline left as-is); inline JSON unchanged (regression-tested).

Verified via dry-run: `sheet --values @file/@-/inline`, `report --contents @file`, missing-file error. New unit tests cover @file, @-, bare-@, missing-file, strict @file, non-@ passthrough.

PeterGuy326:
Folded into #516 for a unified regression pass (combined with #515). Closing this in favor of #516; branch `fix/report-contents-file` kept for history.

PR #515, opened 2026-06-26
fix(compat): hide phantom override commands from --help (tool-existence guard)

PeterGuy326:
## Problem

Override leaves whose backing MCP tool isn't actually deployed still render in `dws <svc> --help`, then fail at invocation with *tool not found*. Measured against the **live open-edition** tools/list (via `cache refresh`): **43 phantom commands across 391 overrides**. Worst offender is attendance — 38 declared overrides, only **4** deployed tools (group_or_rules / record / shift / summary); the other 34 (class/group/vacation/overtime/schedule/approve/check/report/...) are entire undeployed feature areas.

The binary faithfully renders the metadata — there was **no existence guard**: the override loop builds a cobra command unconditionally and only consults tool details to *enrich* descriptions/flags.

## Fix — runtime guard (this PR)

`BuildDynamicCommands` now takes an `existingTools` oracle (CLI slug → live tool-name set, from the **`tools/`** cache). A leaf whose tool is absent from its resolved server's set is marked `Hidden` (still invocable — just off the help surface); groups left with no visible leaf are collapsed.

Safety rails — the naive "skip if tool not in detail set" version would **blank the entire command tree on a cold cache** (the `detail/` partition is empty even after a refresh; supplement/overlay servers have no detail at all):

- acts **only** when a server's tool set is KNOWN and non-empty; absent/empty = unknown = **keep**. Cold cache / overlay / plugin paths pass `nil` → no-op.
- the oracle reads the **`tools/`** partition (populated by `cache refresh`), not `detail/`.
- `serverOverride` leaves resolve against the **target** server's set (contact→hrmregister, doc→doc-comment stay visible).
- `pipeline` leaves are never hidden.

## Publish-time companion

`scripts/dev/check-phantom-overrides.py` diffs envelope `toolOverride` keys against the live tool sets and fails on **un-acknowledged** phantoms (`hidden:true` overrides count as acknowledged). This is the discovery-side (Plan A) gate; the matching envelope change (37 phantom overrides → `hidden:true`, 1 helper-shadowed override removed) ships separately via the discovery-release/Portal flow, not git.

## Verification

- `attendance --help`: **17 → 4** real commands; warm tools cache.
- Empty tools cache: **all 17 stay visible** (guard inert — cold cache never blanks the tree).
- 22 products and healthy services (calendar, report, …) unaffected.
- `phantom_guard_test.go`: hide / cold-cache-keep / empty-or-absent-keep / serverOverride-routing / pipeline-skip. Full `internal/compat` + `internal/app` suites green.

## Scope note

Independent of #514 (report `--contents-file`). No overlap in touched code regions; `report` has zero phantom drift.


PeterGuy326:
Scope grew: this branch now also carries (a) the empty-group collapse so groups left childless by hidden overrides don't render as empty headings, and (b) skill reference alignment (14 files) so every documented command/flag matches the real CLI — attendance.md rewritten around its 4 real commands, contact `label` / ding `message list|receiver-status` phantoms removed, and runtime-failure flag fixes (mail --body→--content, doc pagination, chat send-by-bot --at-*, calendar/minutes/wiki). report.md left to #514.

PeterGuy326:
Folded into #516 for a unified regression pass (combined with #514). Closing this in favor of #516; branch `fix/phantom-command-guard` kept for history.

PR #536, opened 2026-06-30
fix(pat): keep agent code grants aligned

shangguanxuan633-lab:
## Summary

Fix PAT grant/check key divergence for host-provided agent codes.

- keep explicit `DINGTALK_DWS_AGENTCODE` values verbatim on outbound identity headers
- stop synthesizing `custom` into `x-dingtalk-dws-agent-code` / `x-dws-agent-instance-id` for unidentified hosts
- preserve `dws pat chmod --agentCode` as the highest-priority grant target over the env fallback
- update tests and docs for the common cross-host contract

## Root Cause

`pat chmod --agentCode <value>` could grant under one agent key while follow-up normal commands sent a different derived/synthetic agent key in `x-dingtalk-dws-agent-code`. The server then checked a different key and reported PAT authorization again.

## Validation

Local package build:

```sh
VERSION=1.0.46-agentcode DWS_PACKAGE_VERSION=1.0.46-agentcode make package
```

Packaged binary verified:

```text
Version: v1.0.46-agentcode
Commit: 2df7f4e
```

Focused tests:

```sh
go test ./internal/auth ./internal/app ./internal/pat
```

Real PAT/todo authorization loop using packaged binary and existing public env contract:

```sh
DINGTALK_DWS_AGENTCODE=QoderWork /tmp/dws-agentcode-package-test/dws pat chmod todo.task:delete --grant-type permanent --agentCode QoderWork --yes --format json
```

Server returned `agentCode: "QoderWork"` and `grantedScopes: ["todo.task:delete"]`.

Then two self-test todos were created and deleted with the same `DINGTALK_DWS_AGENTCODE=QoderWork` and no second chmod. Both deletes succeeded without another PAT prompt.

## Notes

No new environment variables are introduced. Hosts should use the existing `DINGTALK_DWS_AGENTCODE` contract when they need follow-up commands to use the same agent key as the PAT grant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is inspired by [Keep a Changelog](https://keepachangelog.com/) and th
 
 ## [Unreleased]
 
+### Fixed
+
+- **PAT agentCode grants no longer split from follow-up command checks** (`internal/auth/agent_code_detect.go`, `internal/app/runner.go`, `internal/pat/chmod_test.go`) — explicit `DINGTALK_DWS_AGENTCODE` declarations are now forwarded verbatim as the common cross-host contract, and unknown hosts no longer synthesize `custom` into `x-dingtalk-dws-agent-code` / `x-dws-agent-instance-id`. `pat chmod --agentCode` remains the highest-priority grant target and still wins over the env fallback.
+
 ## [1.0.45] - 2026-06-29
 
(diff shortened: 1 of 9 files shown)


PeterGuy326 (approved):
Reviewed: fix aligns the PAT grant path and the outbound identity-header path onto the same verbatim `AgentCodeFromEnv()` value, so a host-declared agent code (e.g. QoderWork) no longer diverges between grant key and check key. Unknown hosts now stay empty instead of synthesizing `custom`, so no synthetic PAT key is sent. Verified locally: `go test ./internal/auth ./internal/app ./internal/pat` all pass, `go vet` clean, CI green. Note (non-blocking): `normalizeAgentCode` now splits qoder/qoderwork into distinct identities — affected hosts need a one-time re-grant.

PeterGuy326:
你那个 pat agentcode 的 PR 我 review 了 没问题 合进 main 了
麻烦拉下最新主干再整体跑一遍 重点看下声明 QoderWork 的时候 pat chmod 授权和后面普通命令是不是同一把 key 不再重复要 pat
对了 这次把 qoder 和 qoderwork 拆成两个身份了 老的按 qoder 授过权的 host 得重新授一次 这个是预期的 别当成 bug

PR #567, opened 2026-07-07
fix(plugin): accept delivery beta version aliases

PeterGuy326:
## Summary
- Normalize official low-numbered delivery beta CLI tags like `v0.0.48-beta.1` to the `v1.0.48` capability line for plugin `minCLIVersion` checks.
- Preserve normal semver ordering for non-delivery versions and keep the original version string in validation errors.
- Add regression coverage for `v0.0.48-beta.1` against legacy plugin requirements such as `0.2.33` and `1.0.40`.

## Verification
- `go test ./internal/plugin`
- `go test ./internal/app -run 'TestPlugin|TestVersion'`\n- Built with `-X github.com/DingTalk-Real-AI/dingtalk-workspace-cli/internal/app.version=v0.0.48-beta.1`; verified `dws version` and `dws plugin list` no longer emit `plugin requires CLI >= 0.2.33` and `conference` remains enabled.\n- `go test ./...` still fails in `test/scripts` on the post-goreleaser fixture with `tar: Error opening archive: Unrecognized archive format` while ad-hoc signing fake darwin archives; unrelated to this plugin validation change.


PeterGuy326:
Closing as obsolete historical compatibility. This patch only normalizes the one-off `v0.0.48-beta.1` delivery tag; the release line has since moved to normal `v1.x` versions. Merging it now cannot change that already-built historical binary and does not apply to current or future version ordering.

PR #575, opened 2026-07-07
fix: honor global --jq/--fields on product commands; round-2 QA fixes

audanye-sudo:
## Summary

A second full real-machine QA pass (30 agents over ~614 commands, mono/multi skill docs, and helper scripts) on top of v1.0.49. The headline fix makes the global `--jq` / `--fields` output filters actually work for product commands; the rest are the CLI, script, and doc issues the pass surfaced. Genuinely server-side limitations are documented rather than code-patched.

## Global --jq / --fields (main fix)

The filters already worked for `dws api`, but every product (MCP) command printed through a helper `Formatter` that ignored them, so `--jq` / `--fields` were silent no-ops there. This exposes `Fields()` / `JQ()` on the `ToolCaller` interface and applies the existing `output.WriteFiltered` path inside the helper Formatter's `PrintJSON`. Verified on the deployed binary, e.g. `dws contact user get-self --jq '.result[0].orgEmployeeModel.userId'` now returns just the value.

The few bespoke utility commands (`auth` / `config` / `profile` / `doctor` / `schema`) still encode JSON directly and are documented as not honoring these flags yet.

## Other CLI fixes

- **sheet write-image**: emit clean JSON under `--format json` (suppress progress lines that leaked onto stdout, matching media-upload / export).
- **sheet range batch-set-style**: under `--format json`, collect per-item results into a single JSON object instead of printing N separate ones.
- **chat download-media**: create the output directory when missing and strip URL-encoded path separators from the inferred filename, so the file lands instead of failing on a missing subdirectory.
- **--help text**: correct pat chmod scope names, non-existent subcommands, flag requiredness, and an `alxs -> axls` typo across pat/aitable/sheet/chat/attendance.

## Helper scripts (mono and multi)

- `minutes_extract_todos`: parse `dingtalkTodoList` / `actions` (there is no `todos` key), so todos are no longer silently dropped.
- Sync the multi copies of `chat_export_messages` / `chat_history_with_user` (were crashing with `AttributeError`), `minutes_list_parse` / `minutes_recent_summary`, and `calendar_free_slot_finder` to the fixed mono versions.

## Skill docs (mono and multi)

Correct return-structure keys, flag names, deprecated command routing (`doc download` -> `drive download`), enum values, and server-side limitations across products; update the global reference to note `--jq` / `--fields` now apply to product commands.

## Notes

- All fixes verified on a locally built and deployed binary against a test account using self-created test resources only; no production data was touched.
- `go build`, `go vet`, `gofmt`, and the output/helpers/pat test suites pass. One pre-existing app test (`TestPrintPatAuthJSON_MachineReadable`, browser-open in a headless env) fails on clean main as well and is unrelated.
- Not changed by design: server-side behaviors (stringified `success` values, `aggregate` clear, org-restricted meeting rooms / dashboard PUBLIC sharing, roster-permission commands, `dev doc search` missing tool key) are documented, not code-patched.


PeterGuy326 (approved):
Reviewed the PR diff, CI status, and local verification. No blocking issues found.

PR #578, opened 2026-07-08
fix: keep keychain reads side-effect free

LastdianXuan:
## Summary

- make keychain read paths side-effect free so `auth status` and diagnostics do not create a new DEK
- surface explicit `dek_missing` and `keychain_unavailable` diagnostics for auth status / doctor flows
- keep DEK creation on write paths such as token or secret persistence

## Root cause

Read-only keychain paths could fall through to DEK creation. On macOS or `DWS_DISABLE_KEYCHAIN=1` file fallback this meant commands such as `auth status` could create `keychain/dws-cli/dek` even when no login state existed, and missing DEKs were harder to diagnose.

## Validation

- `go test ./internal/keychain -count=1`
- `go test ./internal/app -count=1`
- `go build -o /tmp/dws-main-dek-pr-bin ./cmd`
- empty-config probe with `DWS_DISABLE_KEYCHAIN=1`, `DWS_KEYCHAIN_DIR`, and `DWS_CONFIG_DIR`: `auth status` returns unauthenticated and creates no `dek` file

### DIFF
--- a/internal/app/auth_command.go
+++ b/internal/app/auth_command.go
@@ -1226,7 +1226,17 @@ type authStatusDiagnostic struct {
 }
 
 func authStatusDiagnosticFromError(err error) *authStatusDiagnostic {
-	if err == nil || !keychain.IsUnavailable(err) {
+	if err == nil {
+		return nil
+	}
+	if keychain.IsDEKMissing(err) {
+		return &authStatusDiagnostic{
+			Reason:  "dek_missing",
+			Message: "本地登录密钥缺失，无法解密已保存的登录态",
+			Hint:    "重新登录以生成新的本地登录密钥；如仍异常，可先清理本地登录态后再登录。",
+		}
+	}
+	if !keychain.IsUnavailable(err) {
 		return nil
 	}
 	return &authStatusDiagnostic{
--- a/internal/app/auth_command.go
+++ b/internal/app/auth_command.go
@@ -446,6 +446,7 @@ func newAuthStatusCommand() *cobra.Command {
 			authenticated := false
 			refreshed := false
 			var tokenData *authpkg.TokenData
+			var statusErr error
 			provider := authpkg.NewOAuthProvider(configDir, nil)
 			configureOAuthProviderCompatibility(provider, configDir)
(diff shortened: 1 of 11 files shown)


PeterGuy326 (commented):
## Code Review

整体方向对，核心改动（读/写路径分离 DEK）解决了 `auth status` 弹钥匙串框的根因，测试覆盖充分。两个小点：

**1. `authStatusDiagnosticFromError` 未知错误静默吞掉**

```go
if !keychain.IsUnavailable(err) {
    return nil
}
```

非 keychain 的未知错误会返回 nil diagnostic，目前 OK 因为已知错误只有这两种。但后续如果 auth 链路新增其他错误类型，这里不会有任何诊断输出，排查时容易漏。建议至少 log 一下 unrecognized error，或者返回一个 `reason: "unknown"` 的 fallback diagnostic。

**2. `doctorCheckKeychain` detail 合并可以简化**

```go
detail := make(map[string]string, len(diagnostic.Detail)+1)
for k, v := range diagnostic.Detail {
    detail[k] = v
}
detail["reason"] = diagnostic.Reason
r.Detail = detail
```

可以直接 `maps.Clone(diagnostic.Detail)` + 赋值 reason，或者如果 `Detail` 本身是 map literal 就不需要 copy（因为不会被外部修改）。当前写法没错，只是多了一层手动 copy。

---

另外，多组织多账号场景也一起在这个 PR 做验证：profile 切换后或过一段时间，某些组织的登录态会莫名丢失。帮忙用 `dws profile list` / `dws profile switch` 多切几个组织，分别跑 `auth status` 看登录态是否稳定，有没有切换后丢失的情况。

PeterGuy326 (approved):
Reviewed the DEK read-only change and verified the critical scenarios locally against base vs PR:

- empty `auth status` no longer creates `keychain/dws-cli/dek`
- existing token ciphertext with missing DEK reports `dek_missing` and does not create a replacement DEK

Also checked the multi-profile path locally; the one abnormal profile is reproducible on the base commit and is not introduced by this PR. CI is green. No blocking findings.

PR #597, opened 2026-07-10
fix: preserve macOS auth across keychain mode changes

PeterGuy326:
## Summary

- keep macOS file-DEK credentials readable when a normal process can also access the system Keychain
- preserve the DEK that encrypted an existing token entry, and reject writes when no available DEK can decrypt existing ciphertext
- make new profile token slots follow the canonical `auth-token` backend
- preflight existing token slots before OAuth exchange, device login, or refresh can rotate a server-side refresh token
- add `dws auth migrate-keychain --to file-dek` so existing system-Keychain-backed logins can be moved to the sandbox-compatible backend without exposing tokens
- run the Darwin-only auth/keychain regression suite on a real macOS GitHub runner

## Root cause

`DWS_DISABLE_KEYCHAIN=1` stores the DEK in a file, while the default macOS path stores it in the system Keychain. Both modes previously read and wrote the same `auth-token*.enc` files. If local processes used different values for `DWS_DISABLE_KEYCHAIN`, one process could no longer decrypt the other process's ciphertext and could overwrite it with its own DEK.

## Behavior after this change

- system mode tries the system DEK, then an already-existing file DEK, without creating key material on reads
- updates to existing ciphertext stay encrypted with the DEK that successfully decrypted it
- file-DEK mode cannot overwrite system-Keychain-backed ciphertext
- unknown ciphertext fails closed and remains byte-for-byte unchanged
- OAuth/device login and refresh stop before remote token exchange when a target slot is unreadable
- legacy profile fallback happens only when the selected corp slot is confirmed absent
- `auth status` reports `ciphertext_key_mismatch` instead of treating the state as an ordinary logout

## Safe migration and recovery

For a login that still works in a normal terminal but not in a sandbox:

```bash
env -u DWS_DISABLE_KEYCHAIN dws auth migrate-keychain --to file-dek --dry-run --format json
env -u DWS_DISABLE_KEYCHAIN dws auth migrate-keychain --to file-dek --yes --format json
DWS_DISABLE_KEYCHAIN=1 dws auth status --format json
```

The migration:

- selects only the legacy and profile-scoped auth token entries; unrelated application secrets cannot block login recovery
- decrypts and validates every selected entry before creating a file DEK or staging writes
- re-encrypts all entries in memory, verifies the result, stages every file, then commits with atomic renames
- is serialized with profile/token writes and can be rerun after an interrupted commit
- requires explicit `--yes`; `--dry-run` is side-effect free

If an auth ciphertext is genuinely damaged, the error identifies the profile recovery path. Per-profile logout is preferred; global `auth reset` remains the last resort.

## Validation

- `go test -count=1 ./internal/keychain ./internal/auth ./internal/app`
- `go test -race -count=1 -timeout=10m ./internal/keychain ./internal/auth ./internal/app`
- `go vet ./...`
- `staticcheck ./internal/keychain ./internal/auth ./internal/app`
- `make build`
- `make policy`
- cross-compiled `linux/amd64` and `windows/amd64`
- real-machine migration dry-run against the reported three-profile state: four auth entries validated, zero writes
- isolated compiled-binary E2E still preserves authentication across file-DEK/system-Keychain reads and updates

The full `go test ./...` run still has the pre-existing macOS unsigned-binary execution failure and three release-package tests that require a package version/tag. All packages changed by this PR pass, including race detection.

## Cross-platform CI coverage

- Linux/Ubuntu runs the existing full command/internal suite with race detection, coverage, policy, lint, and edition contracts.
- macOS runs the Darwin-only Keychain/auth regression suite with race detection.
- Windows builds the native CLI and runs the DPAPI/keychain/auth packages plus the new migration-command behavior tests.

A broader Windows internal suite was also attempted and exposed unrelated pre-existing portability failures in auth export test isolation/log cleanup and a Unix 0755 permission assertion in the upgrader. The PR job intentionally remains scoped to the auth/keychain surface changed here while the full native CLI build still protects cross-package compilation.

## Review follow-up

Addressed both P1 boundaries raised during review:

- key selection now receives the account name; only `auth-token` and `auth-token:*` inherit the auth anchor, while newly stored client secrets, app tokens, and app secrets remain protected by the normal system-Keychain DEK after auth migration
- OAuth/device/exchange preflight now validates every file-backed `auth-token*.enc` entry, including orphan profile ciphertext written before an interrupted `profiles.json` update
- migration and preflight share the same deterministic auth-ciphertext inventory boundary
- added regression coverage for post-migration non-auth secret isolation and fail-closed orphan ciphertext detection before any HTTP token exchange

Follow-up validation:

- `go test -count=1 ./internal/keychain ./internal/auth ./internal/app`
- `go test -race -count=1 -timeout=10m ./internal/keychain ./internal/auth ./internal/app`
- `go vet ./...`
- `staticcheck ./internal/keychain ./internal/auth ./internal/app`
- `make build`
- `make policy`
- Linux and Windows CLI/package cross-compilation


shangguanxuan633-lab (changes requested):
发现 2 个阻断问题，暂不能 approve：

1. [P1] `internal/keychain/keychain_darwin.go:351-370,388-402`：`platformSet` 为任意新 account 调用 `keyForNewEntry(service)`，而该函数无条件继承 canonical `auth-token` 的 DEK。执行本 PR 的 token migration 后，auth-token 使用 file-DEK；此后普通 macOS 进程新增的 `client-secret:*`、app token 或 `StoreSecret` 条目也会静默落到 file-DEK，可被 `DWS_DISABLE_KEYCHAIN=1` 沙箱读取。这违反“仅迁移 auth entries / ignores unrelated application secrets”的安全契约，也让非认证密钥从系统 Keychain 保护降级。请让 key selection 接收 account，仅 `auth-token` 与 `auth-token:*` 继承 auth anchor；其它新条目在正常模式继续使用 system Keychain。补迁移后新增 client/app secret 的隔离回归测试。

2. [P1] `internal/auth/keychain_store.go:109-139`：OAuth/device/exchange preflight 只枚举 `profiles.json` 中已登记的 corp slot，遗漏磁盘 orphan `auth-token_*.enc`。`internal/auth/token.go:125-131` 当前先写 profile ciphertext、后更新 profiles.json；两步间中断即可留下 orphan。若 orphan 使用另一 DEK，下一次登录/交换会绕过预检，远端 exchange 完成后才在保存同 corp slot 时失败，违背“远端 token 轮换前 fail closed”。请预检磁盘全部 auth-token ciphertext（迁移代码 `internal/keychain/migration_darwin.go:39-69` 已有相同枚举边界），并补“密文已落盘但 profile 元数据未落盘”的回归测试。

验证：PR 所有 GitHub checks 均通过；本地 `go test -count=1 ./internal/keychain ./internal/auth ./internal/app` 与 `go vet ./internal/keychain ./internal/auth ./internal/app` 通过。上述是现有测试未覆盖的安全/恢复边界。

shangguanxuan633-lab (approved):
## Review verdict: APPROVE

**CI**: Lint / Test / Test (macOS auth/keychain) / Test (Windows) / Coverage / Policy / Edition Contract / Multi Profile E2E = pass.

### Why this is safe to merge
- Fail-closed key collision fix is real: `decryptWithAvailableDEK` does not create keys on read; `platformSet` reuses the DEK that decrypted existing ciphertext; file-DEK cannot overwrite foreign ciphertext (`ErrCiphertextKeyMismatch`).
- Remote token rotation is gated: OAuth / device login / refresh preflight unreadable slots before HTTP exchange.
- Migration is constrained: auth-token* only, validate-all-then-stage, atomic rename, `--yes` required, dry-run side-effect free, serialized via `withProfilesLock`.
- `auth status` surfaces `ciphertext_key_mismatch` instead of ordinary logout.
- Docs/CLI/skills stay consistent; non-Darwin migration fails closed.

### Residual (non-blocking P2)
1. `Exists()` / legacy-only load can mask mismatch as “absent” when no profiles exist.
2. System Keychain `Unavailable` short-circuits before file-DEK fallback (intentionally fail-closed; docs could be sharper).
3. Migration holds plaintext in memory without zeroize; multi-file commit relies on rerun (documented).
4. Windows CI intentionally scoped and without `-race` — acceptable for this surface.

**No blocking findings.** Residual items can be follow-ups; do not block this PR.

shangguanxuan633-lab (approved):
Re-review based on latest head `82abe00`.

Both previously reported P1 findings are resolved:

- `keyForNewEntry(service, account)` now limits auth-anchor DEK inheritance to `auth-token` / profile token accounts; non-auth client/app secrets remain protected by the system Keychain. The regression test verifies they are unreadable in file-DEK-only mode.
- OAuth preflight now validates the full persisted auth-token ciphertext inventory, including orphan slots absent from `profiles.json`; the regression test confirms failure occurs before any HTTP exchange (`calls == 0`).

Validation: all GitHub checks pass; local `go test -count=1 ./internal/keychain ./internal/auth ./internal/app` and `go vet ./internal/keychain ./internal/auth ./internal/app` pass. Cross-platform storage behavior and build-tag implementations were reviewed.

No blocking findings. Approved.

PR #601, opened 2026-07-11
fix(gitee): rewrite release workflow in Gitee Go native format

PeterGuy326:
## Summary

- Rewrite `.workflow/gitee-release.yml` from GitHub Actions syntax to Gitee Go native pipeline format (`triggers.push.tags.prefix`, `step: build@golang`, `inputs.commands`). The old file used `on:`, `jobs:`, `uses: actions/checkout@v4` etc., which Gitee Go cannot execute — so Gitee releases have been stuck at v1.0.44.
- Fix `scripts/dev/build-all.sh` to prefer `sha256sum` (coreutils) over `shasum` (Perl), matching the pattern already used in `post-goreleaser.sh` and `publish-gitee-local.sh`. CentOS (Gitee Go's `build@golang` OS) doesn't have `shasum`.

## After merging — Gitee side setup needed

1. **Enable Gitee Go**: Repo → 管理 → Gitee Go → 开启流水线
2. **Configure secret**: Pipeline settings → 添加参数 → name: `RELEASE_TOKEN`, value: (Gitee personal access token with release perms), mark as secret
3. **Verify**: Push a test tag (e.g. `v1.0.52-beta.0`) and check if the pipeline triggers and publishes

## Test plan

- [ ] Gitee Go pipeline triggers on tag push (`v*`)
- [ ] All 6 platform binaries build successfully on Gitee Go
- [ ] Darwin binaries are ad-hoc signed via `rcodesign`
- [ ] All artifacts upload to the Gitee release page
- [ ] `checksums.txt` is generated with `sha256sum` format

### DIFF
--- a/.workflow/gitee-release.yml
+++ b/.workflow/gitee-release.yml
@@ -1,51 +1,38 @@
-name: Gitee Release
+version: '1.0'
+name: gitee-release
+displayName: 'Gitee Release'
 
-on:
+triggers:
   push:
     tags:
-      - "v*"
-  workflow_dispatch:
-    inputs:
-      version:
-        description: "Release tag to build on Gitee, e.g. v1.0.48"
-        required: false
-        type: string
+      prefix: [v]
 
-jobs:
-  release:
-    runs-on: ubuntu-latest
-    timeout-minutes: 45
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
+stages:
+  - stage:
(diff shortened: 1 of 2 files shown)


PeterGuy326:
Superseded by #620, which consolidates release publication into the guarded beta/stable pipeline and includes the Gitee reconciliation path. Closing this standalone Gitee Go workflow proposal to avoid maintaining a parallel tag-triggered release path. If the sha256sum portability change is still needed, it should be proposed separately.

PR #609, opened 2026-07-13
feat(event): AI-subprocess contract and cobra-synthesized schema

audanye-sudo:
## What

Two related additions to the personal event stack.

### 1. AI-subprocess contract for `dws event consume`

Makes event streaming drivable by an orchestrator without guessing:

- Fixed stderr ready line `[event] ready event_key=<key> bus_pid=<pid>` — block on it, don't `sleep`.
- Final `[event] exited — received N event(s) in Xs (reason: limit|timeout|signal|bus_shutdown)` line; exit code 0 on controlled exit, non-zero (and no `exited` line) on failure.
- stdin-EOF as a graceful shutdown signal, armed only for a parent-controlled pipe stdin on an unbounded run — an interactive TTY and `< /dev/null` never trigger it.
- Ownership-based subscription cleanup: a subscription this run created is unsubscribed on any clean exit; a `--subscribe-id`-reused one is left intact (`--ephemeral` still forces cleanup), so `kill -9` is the only way to leak a server-side subscription.
- Forward `--profile` to the detached bus child so non-default orgs resolve the right credentials, and surface the child's real startup error over the ready pipe instead of an opaque message.

### 2. `dws schema` for event commands

`dws schema "event consume"` (or `event.consume`) returns a machine-readable input schema synthesized from the command's cobra flags, in the same flat shape helper subtrees emit:

```
{ "description", "path", "source": "cobra",
  "parameters": { "<flag>": { "type", "required", "description", "default?" } },
  "arguments": [ { "name", "required", "variadic?" } ] }
```

- Intermediate nodes (`dws schema event`) list their subcommands.
- `source: "cobra"` distinguishes flag-synthesized schema from MCP-fetched (`mcp:<server>`).
- Inherited global flags and hidden internal flags are excluded so the schema describes just that command; visible flags match `--help` exactly.
- Built as a reusable registry (`cobraSchemaRoots`); `event` is the first consumer and more command trees can opt in without further wiring.

## Testing

- `go build`, `go vet`, lint: clean.
- Policy: `check-command-surface --strict`, `check-generated-drift`, `check-open-source-assets`: all ok.
- Unit tests: new `cobra_schema` suite (leaf shape / positional args / dot+space path equivalence / group browse / unknown subcommand / non-registered decline) plus the event contract tests (ready marker, exit reason, stdin-EOF, spawn error surfacing, profile forwarding) pass; `cli` / `contract` / `integration/extensions` suites green.
- The `test/scripts` goreleaser packaging tests fail locally due to a missing real goreleaser dist (macOS ad-hoc signing on fixture tarballs) — unrelated to this change and reproduced identically on a clean tree; they run against real artifacts in CI.

## Docs

- mono + `dingtalk-event` skills document the subprocess contract and clarify the two schema surfaces (`event schema <event_key>` for event output fields vs `dws schema "event consume"` for command input).
- Design notes in `docs/event-subprocess-contract.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,13 @@ The format is inspired by [Keep a Changelog](https://keepachangelog.com/) and th
 
 ## [Unreleased]
 
+### Changed
+
+- **`event consume` AI-subprocess contract** — aligns personal event streaming with the contract an orchestrator can drive without guessing: a fixed stderr ready line `[event] ready event_key=<key> bus_pid=<pid>` (block on it, don't `sleep`); a final `[event] exited — received N event(s) in Xs (reason: limit|timeout|signal|bus_shutdown)` line with exit code 0 on controlled exit and non-zero (no `exited` line) on failure; stdin-EOF as a graceful shutdown signal, armed only for a parent-controlled pipe stdin on an unbounded run (an interactive TTY and `< /dev/null` never trigger it), with a self-explaining diagnostic when it fires; and ownership-based subscription cleanup — a subscription this run created is unsubscribed on any clean exit while a `--subscribe-id`-reused one is left intact (`--ephemeral` still forces cleanup), so `kill -9` is the only way to leak a server-side subscription. Skill docs (mono + `dingtalk-event`) document the contract; design notes in `docs/event-subprocess-contract.md`.
+
 ### Added
 
+- **`dws schema` for registered local commands** — `dws schema "event consume"` (or `event.consume`) now returns a machine-readable input schema synthesized from the command's cobra flags, in the same flat shape helper subtrees emit: `{description, path, source, parameters{<flag>:{type, required, description, default?}}}` plus an `arguments` array for positional inputs, with `source: "cobra"` distinguishing flag-synthesized schema from MCP-fetched (`mcp:<server>`). Intermediate nodes (`dws schema event`) list their subcommands. The mechanism is a reusable registry (`cobraSchemaRoots`); `event` is the first consumer and more command trees can opt in without further wiring. Inherited global flags and hidden internal flags are excluded so the schema describes just that command.
 - **Safe macOS Keychain → file-DEK migration** — `dws auth migrate-keychain --to file-dek` preflights every legacy/profile auth entry before rewriting, ignores unrelated application secrets, supports side-effect-free `--dry-run`, requires explicit `--yes`, and lets sandboxed and normal processes share an existing login without exposing tokens.
 
 ### Fixed
--- a/docs/event-subprocess-contract.md
+++ b/docs/event-subprocess-contract.md
@@ -1,0 +1,115 @@
+# Event consume — AI subprocess contract
+
+Aligns `dws event consume` with the "AI subprocess contract" that
+`lark-cli event consume` exposes, so any orchestrator (Claude Code's
+Monitor, a bash bridge, systemd, an agent plugin) can drive it with zero
+ambiguity: know when it is ready, stop it cleanly, and machine-read why it
+exited.
+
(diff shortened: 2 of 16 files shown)


PeterGuy326:
## Code Review

### Overall

Solid work — the AI-subprocess contract is well-designed and aligns cleanly with lark-cli's approach. The context layering (parentCtx → timeoutCtx → runCtx) for exit reason classification is particularly clean. Test coverage is comprehensive (T1–T4 paths all covered). CI is fully green.

### Findings

#### 1. Minor: stdin-EOF diagnostic bypasses `--quiet`

`internal/event/consume/run.go` — `watchStdinEOF`:

```go
fmt.Fprintln(errOut, "[event] stdin closed — shutting down. "+
    "consume treats stdin EOF as an exit signal ...")
```

The ready marker and bus diagnostic lines are gated on `!cfg.Quiet`, but this diagnostic always writes to stderr. If the parent process is running with `--quiet`, it probably doesn't want a multi-line explanation on stderr either. Consider gating this on quiet as well, or passing a boolean flag to `watchStdinEOF`.

#### 2. Minor: `shouldWatchStdinEOF` hardcodes `os.Stdin.Stat()`

`internal/app/event_command.go`:

```go
fi, err := os.Stdin.Stat()
```

The function checks `os.Stdin` directly while `cfg.Stdin` is set from `c.InOrStdin()`. In practice these resolve to the same fd, but it makes the gating logic impossible to unit-test with an injected reader. Consider accepting an `*os.File` parameter or extracting the stat check, so the decision logic can be tested without relying on the real stdin.

#### 3. Nit: `classifyCancel` timeout-vs-signal priority

When both `--duration` and a signal fire in close succession, `classifyCancel` always returns `"timeout"` because it checks `timeoutCtx.Err()` before `parentCtx.Err()`:

```go
if timeoutCtx != nil && errors.Is(timeoutCtx.Err(), context.DeadlineExceeded) && parentCtx.Err() == nil {
    return "timeout"
}
return "signal"
```

This is arguably correct (the run was bounded anyway), but a user who Ctrl-C's at T=4.9s of a 5s duration might expect `reason: signal`. Consider checking `parentCtx.Err() != nil` first so explicit user intent wins over an incidental deadline. Low priority.

#### 4. Suggestion: goroutine lifecycle for `watchStdinEOF`

The goroutine launched for `watchStdinEOF` has no wait mechanism — `Run()` can return while it's still blocking on `r.Read(buf)`. In practice the goroutine will exit when the process exits or the reader is closed, but for library-style usage (tests, embedded callers) this is a goroutine leak window. Consider adding a `done` channel or `sync.WaitGroup` and waiting on it in the deferred cleanup.

### What looks good

- **Ownership-based subscription cleanup** — `selfCreated` check is the right semantic; matches lark-cli and eliminates the most common subscription leak path.
- **Profile forwarding to bus child** — fixes a real multi-org pain point with a clean `--profile <corpId>` pass-through. Good regression test too.
- **Error surfacing on ready pipe** — `'E'` byte + trailing error text is a simple and effective protocol extension. The 4096-byte limit is sensible.
- **`cobra_schema.go`** — clean design, well-separated concerns, good use of `LocalFlags()` to exclude inherited globals. The `cobraSchemaRoots` registry makes it trivial to onboard more command trees.
- **Test coverage** — 7 new test functions covering all contract paths (ready marker, exit reasons, stdin EOF, profile forwarding, schema synthesis, unknown subcommand, non-registered decline).

### Verdict

No blockers. The findings above are minor improvements to consider. The core contract design is sound, well-tested, and CI is clean. 👍

PeterGuy326 (approved):
LGTM — contract design solid, test coverage comprehensive, CI clean. Minor nits noted in prior comment, non-blocking.

PR #610, opened 2026-07-13
feat(pat): add --all, single-scope revoke, and non-interactive confirmation

shangguanxuan633-lab:
## Summary

- Keep the existing `dws pat chmod` entry point and add the smallest command-surface extension: grant-only `--all`, single-scope boolean `--revoke`, and completed PAT-specific `--yes` behavior.
- `--revoke` accepts exactly one positional scope. Its `--dry-run` is entirely local, while the confirmed write calls only the dedicated `pat.scope_revoke` tool. This CLI does not introduce or call `pat.batch_revoke`.
- Keep all-scope grants plan-first through `pat.batch_plan` -> `pat.batch_grant`, require explicit `--yes` for writes, and fail closed when the connected server does not attest the new protocol.
- Make PAT `--yes` fully non-interactive: pending authorization returns the structured action, flow, URI, and trace contract without opening a browser, polling, persisting an interactive flow, or retrying automatically.
- Allow global `--dry-run` to perform exactly one read-only `pat.batch_plan` call while every grant, revoke, and unrelated tool remains locally short-circuited.
- Update README, changelog, and bundled mono/multi skill guidance without adding another authorization command.

## Safety and compatibility

- Revoke rejects zero or multiple scopes and conflicts with `--all`, `--recommend`, product/domain selectors, `--grant-type`, and `--session-id` before any MCP call.
- The mutation payload is exactly `{scope, agentCode?}`. There is no revoke plan, scope expansion, batch fallback, or legacy grant fallback.
- An explicit `agentCode` is injected into the trusted request context and must be attested exactly in the response. The response must also attest `success=true`, `code=OK`, `operation=revoke`, the exact requested `scope`, and `grantPolicy=ONCE`; otherwise the CLI fails closed.
- `--all` remains grant-only and never falls back to the legacy single-grant tool.
- The interaction model references the official Lark CLI v1.0.68 auth surfaces: [`auth login`](https://github.com/larksuite/cli/blob/v1.0.68/cmd/auth/login.go), [`auth logout`](https://github.com/larksuite/cli/blob/v1.0.68/cmd/auth/logout.go), shared [`--yes` behavior](https://github.com/larksuite/cli/blob/v1.0.68/shortcuts/common/runner.go#L967-L973), and token [`revoke`](https://github.com/larksuite/cli/blob/v1.0.68/internal/auth/revoke.go). This PR reuses the explicit `--all` / confirmation pattern while keeping PAT grant removal separate from OAuth logout.

## Verification

- [x] `go test ./internal/pat -count=1`
- [x] `go test -race -count=1 ./internal/pat ./internal/app ./internal/executor`
- [x] `make build`
- [x] `make test`
- [x] `make policy`
- [x] `make edition-test`
- [x] `git diff --check`
- [x] zero `pat.batch_revoke` / `batch_revoke` references
- [x] isolated Darwin/arm64 release-style package: Mach-O arm64, ad-hoc codesign verification, archive checksum, command-surface smoke, and current-account auth/search checks
- [x] one real chat smoke from the packaged binary to the uniquely resolved test recipient; send status `SUCCESS` with a fixed idempotency UUID and no retry
- [x] GitHub CI: Lint, Test, macOS auth/keychain, Windows, Coverage, Policy, Edition Contract Tests, and Multi Profile E2E

`go test ./...` reaches one unchanged environment/baseline failure: `TestStdioClientEndToEnd` times out during its 10-second initialize step. The PAT packages pass, the affected stdio path is unchanged from `upstream/main`, and all repository build/policy/edition gates above pass. A separate unsigned-Darwin test that was killed under concurrent memory pressure passed on isolated rerun.

## Backend dependency: single-scope revoke

This CLI intentionally does **not** introduce or call `pat.batch_revoke`. Revoke depends on a small Gateway adapter before it can succeed end to end:

- [ ] Register a PAT-only synthetic tool named `pat.scope_revoke` with one required non-blank scalar argument, `scope`, plus optional `agentCode`. Do not accept `--all`, product selectors, multiple scopes, `grantType`, or client-supplied `orgId` / `uid`.
- [ ] Reuse the trusted PAT identity, access-key, CLI/DWS authorization, security-check, and write-caller allowlist gates. Inject `orgId` and `uid` from trusted context; never trust caller-supplied identity.
- [ ] Route through explicit `plan` / `grant` / `scope_revoke` / unknown-error branches. Never let an unknown synthetic tool fall through to `batchGrantAuth`.
- [ ] Before mutation, query the exact scope policy. If the explicit policy is `DENIED`, return `PAT_SCOPE_NOT_GRANTED` and do not mutate: Core's `ONCE` behavior also clears a deny, which could otherwise increase privilege. Blank or unknown scopes must fail closed.
- [ ] Execute only `com.dingtalk.pat.hsf.PatAuthService:1.0.0#updateScopeGrantPolicy(UpdateScopeGrantPolicyRequest)` with trusted `orgId`, `uid`, resolved/delegated `agentCode`, the exact scope, and Gateway-fixed `grantPolicy="ONCE"`.
- [ ] Normalize `ServiceResult<Void>` correctly: `success=true` with `result=null` is a valid success and must not use the existing batch-result normalizer.
- [ ] Return `data.operation="revoke"`, `data.scope=<exact request scope>`, `data.grantPolicy="ONCE"`, and the exact resolved `data.agentCode` when one was requested. Do not claim a mutation count or `revoked=true`, because Core treats missing/already-revoked rows as idempotent success.
- [ ] Fence or cancel older pending authorization flows for the same identity/scope so an earlier approval cannot restore a grant after revoke.
- [ ] Add tests proving exactly one `updateScopeGrantPolicy(ONCE)` call, zero `batchGrantAuth` calls, DENIED never reaches mutation, trusted identity wins, `ServiceResult<Void>` succeeds, malformed/misrouted responses fail, and unknown synthetic tools never fall through to grant.

`--dry-run` remains a CLI-local preview and performs zero Gateway/HSF calls. The current Gateway does not expose `pat.scope_revoke`, so this PR should remain draft until that adapter and the required Core behavior are deployed.

## Intentional scope cuts

- No batch revoke and no revoke-all behavior.
- No new `pat revoke` subcommand.
- No OAuth logout/token-revoke behavior.
- No Gateway/Core changes in this repository.
- No live PAT revoke was attempted during CLI smoke testing; the real side-effect smoke exercised packaged-binary account, lookup, and message delivery only.


shangguanxuan633-lab:
Superseded by #614, which is rebased onto the latest main, squashed to one clean commit, and includes the matching Core/Gateway server contracts plus final dry-run/revoke safety fixes.

PR #635, opened 2026-07-15
fix(auth): refresh rejected tokens and retry once

shangguanxuan633-lab:
## Summary

- consume `AuthRefreshRequired` from both transport auth errors and MCP tool-result classification
- atomically refresh the exact access token rejected by the server, reusing a token already rotated by another goroutine or process
- retry only the current endpoint and invocation once, without replaying multi-profile dispatch
- conditionally remove only the current profile credential after refresh failure or retry exhaustion
- invalidate core and edition-owned token caches before retrying
- keep ordinary HTTP 403, PAT scope, permission, and business errors out of the refresh path

## Root cause

The Wukong overlay already returned an `AuthRefreshRequired` marker for explicit token failures, but the core runner treated that marker as a terminal business error. The previous refresh helper had also been removed, so no refresh, retry, or credential cleanup happened.

## Safety

- refresh and rejected-token comparison share the existing process and file lock
- cleanup uses compare-and-delete so a stale request cannot remove a newly rotated credential
- explicit `--token` and plugin-auth invocations are not auto-refreshed
- logs include structured recovery reasons but never token values

## Validation

- `go test -race ./internal/auth -run 'TestForceRefreshRejectedToken|TestDeleteTokenDataIfAccessTokenMatches' -count=1`
- `go test -race ./internal/app -run 'TestRuntimeRunnerAuthRefresh|TestAuthRefresh' -count=1`
- `go test ./internal/app ./internal/auth ./pkg/authretry -count=1`
- `go test ./...`
- `git diff --check`

## Downstream integration

Companion Wukong CR: https://code.alibaba-inc.com/dingtalk-workspace/dws-wukong/codereview/28610386

It registers the new `InvalidateAuthCaches` hook, narrows generic auth-service classification, and adds the cache/classification regression tests. After this PR is merged and tagged, Wukong still needs to bump `CLI_UPSTREAM_TAG` as a release step.

### DIFF
--- a/internal/app/auth_refresh_retry.go
+++ b/internal/app/auth_refresh_retry.go
@@ -15,6 +15,16 @@ package app
 
 import (
 	"context"
+	"errors"
+	"log/slog"
+	"net/http"
+	"strings"
+
+	authpkg "github.com/DingTalk-Real-AI/dingtalk-workspace-cli/internal/auth"
+	apperrors "github.com/DingTalk-Real-AI/dingtalk-workspace-cli/internal/errors"
+	"github.com/DingTalk-Real-AI/dingtalk-workspace-cli/internal/executor"
+	"github.com/DingTalk-Real-AI/dingtalk-workspace-cli/pkg/authretry"
+	"github.com/DingTalk-Real-AI/dingtalk-workspace-cli/pkg/edition"
 )
 
 // authRetryingKey marks a context that has already attempted one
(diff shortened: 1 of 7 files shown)


PeterGuy326:
Closing as superseded by #687. The newer PR explicitly carries forward this rejected-token refresh/retry work and expands it to the complete long-lived token, PAT veto, CAS cleanup, diagnostics, and event-consumer model. Keeping both open would duplicate and conflict on the same auth paths.

PR #658, opened 2026-07-17
fix(chat): use openconversation_id for group message lists

LastdianXuan:
## What changed

- Send `openconversation_id` to `list_conversation_message_v2`.
- Apply the same mapping to `chat message list`, `chat +messages-list`, and `chat +chat-messages`.
- Add regression assertions that legacy `openCid` and `cid` are not sent.

## Root cause and impact

The CLI accepted the group ID through `--group` and its aliases, but forwarded it under legacy argument names. The current Open MCP metadata requires `openconversation_id`, so real group-message queries could not resolve the conversation correctly.

## Validation

- `go test ./... -count=1`
- `go vet ./...`
- `make build`
- `make policy`
- Native changed-code coverage: 100.0% (10 executable statements; target 80.0%)

### DIFF
--- a/internal/helpers/chat.go
+++ b/internal/helpers/chat.go
@@ -1351,10 +1351,9 @@ func newChatCommand() *cobra.Command {
 			timeVal := mustGetFlag(cmd, "time")
 			if groupID != "" {
 				toolArgs := map[string]any{
-					"openCid": groupID,
-					"cid":     groupID,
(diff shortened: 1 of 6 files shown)


LastdianXuan:
Superseded by #662, which combines the confirmed v1.0.53 evaluation fixes into a single review unit.

PR #660, opened 2026-07-17
fix(audit): emit structured verify results in JSON mode

LastdianXuan:
## What changed

- Emit a single structured payload from `audit verify --format json`.
- Include `valid`, `file`, `brokenAt`, and the failure `reason` when the hash chain is broken.
- Preserve the non-zero exit behavior for invalid chains and keep human output unchanged.
- Add regression coverage for valid and broken chains.

## Root cause and impact

`audit verify` printed human-oriented status directly, even when JSON output was requested. Machine callers therefore could not reliably parse verification results or inspect the break position and reason before the command exited.

## Validation

- `go test ./... -count=1`
- `go test ./internal/app -count=1`
- `go vet ./...`
- `make build`
- `make policy`
- Native changed-code coverage: 85.7% (14 executable statements; target 80.0%)

### DIFF
--- a/internal/app/audit_command.go
+++ b/internal/app/audit_command.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/DingTalk-Real-AI/dingtalk-workspace-cli/internal/audit"
+	"github.com/DingTalk-Real-AI/dingtalk-workspace-cli/internal/output"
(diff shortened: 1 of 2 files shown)


LastdianXuan:
Superseded by #662, which combines the confirmed v1.0.53 evaluation fixes into a single review unit.

PR #649, opened 2026-07-16
fix(release): add fast guarded release and recovery paths

PeterGuy326:
## Summary

- Reuse a six-hour preflight proof bound to the exact version, commit, repository, branch, and beta/stable baseline, so a reviewed publish does not repeat tests and packaging.
- Verify immutable-release governance through the same protected workflow identity used by tag publication, while still rechecking CI, delivered baselines, remote main, and tag authority before allocation.
- Add `dws-release recover <version>` for a failed existing annotated tag. Recovery uses the normal build/sign/publish graph behind a reviewed `release-recovery` environment and never moves or recreates the tag.
- Make recovery resumable after GitHub Release sealing by binding reuse to the same workflow run, tag object, commit, notes, and byte-identical finalized artifacts.
- Isolate source installer tests from the real repository binary and raise the full-suite per-package timeout from 120 seconds to 5 minutes after reproducing the script-suite timeout under parallel load.

## Why

The v1.0.53-beta.2 release exposed two operational gaps: local governance credentials could disagree with the workflow identity, and a valid tag whose workflow failed had no safe, reviewed way to resume. The result was pressure to bypass quality gates during an incident. This change keeps the quality contract, removes duplicate expensive work, and adds a narrowly scoped break-glass path for both beta and stable releases.

## Verification

- [x] macOS safe build equivalent to `make build` (`dws-safe-build`; ad-hoc signature verified)
- [ ] `make lint` — blocked by existing staticcheck findings in unmodified files (for example `internal/keychain/keychain_darwin.go` and coverage-only tests)
- [x] `make test` (5 passed suites, 0 failed; recovery suite absent/skipped as before)
- [x] `DWS_PACKAGE_VERSION=0.0.0-test go test ./...`
- [x] `make policy`
- [x] `./scripts/policy/check-generated-drift.sh` (via `make policy`)
- [x] `./scripts/policy/check-schema-catalog.sh` (via `make policy`)
- [x] `./scripts/policy/check-command-surface.sh --strict` (via `make policy`)
- [x] `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.12 .github/workflows/release.yml`
- [x] shell syntax, YAML parse, `git diff --check`, and focused release/recovery regression tests

## Notes

Before enabling recovery, repository administrators must:

- configure `RELEASE_GOVERNANCE_TOKEN` with repository `Administration: read`;
[...]

PR #663, opened 2026-07-17
fix(release): reuse sealed local preflight

PeterGuy326:
## Summary

- Seal a successful local release preflight in worktree-private Git metadata using the exact commit, version, channel, release repository, stable baseline, artifact directory, and SHA-256 of `checksums.txt`.
- Reuse the sealed `dist/` during the matching `--publish` invocation instead of rerunning tests, policy, six-platform packaging, and package-manager install checks. Remote state, CI/publication authority, the release contract, the exact artifact set, embedded versions, and checksums are still revalidated before tag allocation.
- Invalidate the seal and run the full preflight when its source/authority/artifact facts no longer match; remove it after a successful tag push.
- Make the local gate deterministic by rebuilding the root `dws` after tests, preserving any pre-existing root binary across source-install tests, and allowing slow script suites up to five minutes.
- Replace the CI immutable-release check that incorrectly used the built-in `GITHUB_TOKEN` with one shared local/CI verifier. The workflow now prefers `RELEASE_GOVERNANCE_TOKEN`, falls back to the existing admin-owned `HOMEBREW_PR_TOKEN`, and exposes a five-minute read-only `verify_release_governance` probe that creates no tag or package.
- Return actionable diagnostics for missing `Administration: read` permission instead of the opaque `Resource not accessible by integration` failure seen in v1.0.53-beta.3.

## Verification

- [x] `make build`
- [ ] `make lint` (blocked by pre-existing staticcheck findings on current `main` in unrelated coverage/keychain/transport files; none of those files are changed here)
- [x] `make test` before the final workflow-only follow-up; affected release tests rerun below
- [x] `make policy`
- [x] `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.12`
- [x] `go test -timeout 10m -count=1 ./test/scripts` (312.824s)
- [x] `DWS_PACKAGE_VERSION=0.0.0-test go test -timeout 10m -count=1 ./...` before the workflow follow-up; during the final rerun, all non-script packages passed and the duplicate concurrently-running script suite exhausted the aggregate 10-minute limit, then passed in isolation above
- [x] `scripts/release/verify-immutable-release-governance.sh DingTalk-Real-AI/dingtalk-workspace-cli`
- [x] Targeted governance regressions: enabled, disabled, and HTTP 403 permission paths
- [x] Targeted release regressions: check→publish runs each expensive gate once; artifact tampering invalidates the seal and reruns all gates; existing root binary content/mode is restored

## Notes

- No CHANGELOG entry: this changes maintainer-only release infrastructure, not delivered CLI behavior.
- The seal is local cache metadata, not a substitute for remote CI or GitHub release governance. A missing or stale seal falls back to the complete release preflight.
- GitHub documents that the immutable-release settings endpoint requires repository `Administration: read`; the normal Actions permission set does not expose that permission to the built-in token.
- After merge, run the manual `verify_release_governance=true` probe once to prove the configured secret before the next release tag.


PeterGuy326:
Closing as obsolete/superseded by the merged release hardening in #649, #665, #667, #675, and #678. Current main has stricter exact-proof, governance-token separation, Code Admission, repair, and immutable-publication contracts. This branch also falls back from RELEASE_GOVERNANCE_TOKEN to HOMEBREW_PR_TOKEN, which conflicts with the current least-privilege requirement that those credentials remain separate. Any residual seal-hash optimization should be proposed later as a narrow follow-up against current main rather than reviving this mixed, conflicting branch.

PR #694, opened 2026-07-20
Revert "fix(auth): unify token resolution and recover rejected tokens"

sczheng189:
Reverts DingTalk-Real-AI/dingtalk-workspace-cli#687


PeterGuy326:
Closing because the reverted #687 change remains intentionally present on main and is now depended on by later merged auth/event recovery work, including #698. This revert has no recorded regression rationale and now conflicts with main; merging it would remove the current token-resolution and rejected-token recovery contract and its tests.

PR #685, opened 2026-07-19
test(auth): isolate OAuth coverage fixtures

shangguanxuan633-lab:
## Summary

- isolate OAuth coverage fixtures from the macOS system Keychain
- cancel and join every background login goroutine
- replace unbounded channel waits with bounded diagnostics
- keep the fix test-only; production code is unchanged

## Root cause

The coverage-only OAuth tests could enter the real macOS Keychain preflight. A callback timeout then left an unjoined login goroutine alive, which restored package-level hooks underneath later tests and could hang the platform coverage gate.

## Verification

- `go test -count=3 -timeout=3m -run '^TestCrossPlatformCoverageOAuth' ./internal/auth`
- `go test -count=3 -timeout=3m -run '^TestCrossPlatformCoverage' -coverpkg=./internal/app,./internal/auth,./internal/cli,./internal/executor,./internal/pat,./pkg/edition -coverprofile=/private/tmp/auth-crossplatform-fixed.cover -covermode=atomic ./internal/auth`
- `go test -race -count=1 -timeout=5m ./internal/auth`

All passed locally on macOS.

PR #695, opened 2026-07-20
Revert "test(auth): isolate OAuth coverage fixtures"

sczheng189:
Reverts DingTalk-Real-AI/dingtalk-workspace-cli#685


PeterGuy326:
Closing because #685 is an approved test-harness hardening change that remains in active use on main. Reverting it would restore real Keychain exposure, unbounded waits, and unjoined OAuth login goroutines, with no documented regression motivating that rollback.

PR #785, opened 2026-07-24
test(markdown): make disappearing-upload-file cases deterministic

PeterGuy326:
## Summary

- The three `temporary_file_disappears_before_stat` subtests in `markdown_ci_coverage_test.go` raced a polling goroutine against a 16MB write, expecting the removal to land between write and stat. On fast Linux CI runners the window is routinely missed, so the command succeeds and the test fails.
- This fails Coverage jobs on `main` itself (see run on `eaf53bc2`) and blocks every open PR (e.g. #708 hit `TestMarkdownCICoveragePatchEdges` then `TestMarkdownCICoverageOverwriteEdges` on rerun).
- Fix: add an injectable `markdownStatUpload` seam (defaults to `os.Stat`) used by the three upload stat sites; tests override it to remove the file deterministically before delegating to the real stat. Removes the goroutine remover and the 16MB payloads.

## Validation

- `DWS_PACKAGE_VERSION=0.0.0-test go test ./internal/helpers -run 'TestMarkdownCICoverage' -count=5` — PASS, deterministic (was intermittent).
- `go test ./internal/helpers -race -run 'Markdown'` — PASS.
- `go test ./internal/helpers -count=1` (full package) — PASS.
- `gofmt` / `go vet` clean.

## CHANGELOG

Omitted: test-only change plus an internal seam variable; no user-visible behavior change.

### DIFF
--- a/internal/helpers/markdown.go
+++ b/internal/helpers/markdown.go
@@ -27,6 +27,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// markdownStatUpload stats the prepared upload file. Tests override it to
+// deterministically simulate the file disappearing between write and stat.
+var markdownStatUpload = os.Stat
+
 func newMarkdownCommand() *cobra.Command {
 	root := &cobra.Command{
 		Use:   "markdown",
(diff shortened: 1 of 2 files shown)


PeterGuy326:
Superseded: main 02f66df5 已包含等价修复（markdownUploadStat seam + 确定性测试），本 PR 重复，关闭。

PR #451, opened 2026-06-10
feat: support @ mentions in chat message reply

mvanhorn:
## Summary

`dws chat message reply` now accepts `--at-open-dingtalk-ids <csv>` and `--at-all`, forwarding `atOpenDingTalkIds` / `isAtAll` to the upstream `send_personal_message` tool. The reply content goes through the same `marshalMessageContent` path as `chat message send`, so `<@...>` placeholders reach DingTalk clients unescaped. Flag help, both skill reference docs, and CHANGELOG.md document that `--text` must contain matching `<@openDingTalkId>` / `<@all>` placeholders.

This closes the gap in #359: an agent that polls `chat message list-mentions` and replies in-thread had no way to notify the original sender, even though the upstream tool schema already supports these fields (confirmed in #310). The implementation follows the landing plan PeterGuy326 laid out in the issue thread: pure CLI passthrough mirroring the send command's flag names and `splitCSVStrings` handling, with no changes to the bot/webhook send paths.

## Verification

- [x] `make build`
- [x] `make lint`
- [x] `make test`
- [x] `make policy`
- [x] `./scripts/policy/check-generated-drift.sh`
- [x] `./scripts/policy/check-command-surface.sh --strict` (command surface changed: two new reply flags)

## Notes

Table-driven tests cover ids-only, at-all, no-flags (params identical to current behavior), and whitespace-only ids. On `make test`, the cli, contract, and mock_mcp suites pass; the integration/extensions, integration/recovery, and scripts suites fail identically on a clean upstream/main checkout in this environment, so those failures are pre-existing and unrelated to this change. No risks beyond the new optional flags; behavior without them is unchanged.

Fixes #359

PR #794, opened 2026-07-25
feat(chat): support --at-users on chat message reply

### LINKED ISSUE #359: [Feature] 期望 dws chat message reply 支持 --at-users
## Problem Statement

回复消息时没办法 at 指定用户，导致不能及时通知被回复消息的发送者或其他需要关注该回复消息用户。

## Proposed Solution

`dws chat message reply` 支持 `--at-users` 参数，与 `dws chat message send` 命令中的 `--at-users` 使用方式一样。

## Use Case

通过 `dws chat message list-mentions` 定时拉取 @我 的消息后，使用 AI Agent 分析并处理完相关任务后，需要针对指定消息回复对应的消息发送者，为了体现上下文选择用 `reply` 而不是 `send` 命令，并且可以在回复的同时直接 at 被回复消息的发送者。

mvanhorn:
Closes #359. Supersedes #451.

@PeterGuy326 关闭 #451 时说明：需求本身仍有效，但那份实现基于 #565 之前的
executor/helper command 架构，而 `reply` 现已改为静态 handler，文件与主干大幅漂移，
且未覆盖当前 Schema contract，并欢迎基于最新 main 提交新的聚焦 PR。这就是那份新 PR。

## 改动

`dws chat message reply` 新增 `--at-users`，接受逗号分隔的 userId 或
openDingTalkId：

- userId 通过既有的 `resolveOpenDingTalkIDs` 解析为 openDingTalkId；已是
  openDingTalkId 的值原样保留。
- 回复正文中对应的 @ 标识由 `normalizeAtPlaceholders(wrapAngle=true)` 统一为
  `<@openDingTalkId>`，与 `message send` 处理用户身份消息的方式一致。
- 解析结果以 `atOpenDingTalkIds` 随工具参数下发。

未引入新的辅助函数，全部复用 `message send` 已有的 at-mention 路径。

## Schema contract

这是 #451 缺失的部分，本 PR 一并更新：

- `internal/cli/schema_catalog.json`
- `internal/cli/schema_parameter_bindings.json`
- `internal/cli/schema_agent_metadata/chat.json` 与 `index.json`
- `internal/cli/schema_agent_metadata_audit.json`
- `internal/cli/schema_hints/selection/chat.json`

## 文档与测试

- `skills/mono/references/products/chat.md` 与
  `skills/multi/dingtalk-chat/references/chat.md` 同步说明与示例。
- CHANGELOG 条目。
- 新增 `internal/helpers/chat_message_reply_at_users_test.go`。

## 验证

`go build ./...`、`go vet ./...` 通过；`internal/helpers`、`internal/app`、
`internal/cli` 全部通过。


mvanhorn:
Rebased onto main and got CI green -- a1f7dfa + fda8cf2.

The branch was 31 commits behind, which is what all four red checks came down to. `Test (workflow and release contracts)` was failing on `TestReleaseWorkflowWaitsForNPMDistTagPropagation/stale_beta_never_converges`, in `test/scripts/package_script_test.go` -- a file this PR doesn't touch. That test passes on `main`, and it passes here after the rebase; `Coverage`, `Policy` and `Test` were all reporting the same underlying failure.

The rebase conflicted only on generated artifacts (`schema_catalog.json`, `schema_agent_metadata/index.json`, `schema_agent_metadata_audit.json`), all on `source_hash` fields. Rather than hand-merge hashes I took main's copies and re-ran `make generate-schema-agent-metadata` + `make generate-schema-catalog`.

That left the coverage gate at 97.4359% on changed code against a 100% target -- 38 of 39 statements. The one gap was the branch that reports an unresolvable `--at-users` value. `fda8cf2` adds a case where the directory lookup returns no mapping, asserting the command errors and doesn't send a reply containing an unresolved `@`. `chat.go` now has no uncovered blocks.

Everything is green.

mvanhorn:
Merged main. The branch is mergeable again and the only red check is a macOS timeout that also happens on main.

What the merge needed, since main moved a fair way:

- The generated schema artifacts were not merged textually. I took main's copies and reran `make generate-schema`, so `schema_agent_metadata`, the audit file, the catalog and `param_aliases_generated.go` are all generator output rather than hand-resolved JSON.
- `schema_catalog.json` was deleted upstream. I accepted the deletion; the same target regenerates it.
- `schema_parameter_bindings.json` keeps the reply `at-users` -> `atOpenDingTalkIds` mapping, and the reviewed baseline hash is updated to the resulting manifest. That mapping is what `TestChatSchemaSeparatesSendAndReply` asserts, so it fails loudly if the binding is ever dropped again.
- The chat docs were restructured on main from one file into a routing layer with details under `references/chat/`. The `--at-users` documentation went into `references/chat/chat-message.md` rather than into the old monolith.

**On the failing check.** `Test (macOS auth/keychain)` runs

```
go test -v -race -count=1 -timeout=10m ./internal/keychain ./internal/auth ./internal/app
```

and `internal/app` hit `panic: test timed out after 10m0s`. It is not an assertion failure and I do not think it is this PR. Measured on the same machine, same command:

```
this branch   internal/app -race   470.9s
main alone    internal/app -race   552.4s
```

So main is already at roughly 92 percent of the 10 minute budget by itself, and this branch is not slower than main. A GitHub macOS runner is slower than my laptop, which is enough to push it over.

Locally the full job command passes: `go test -race -count=1 ./internal/keychain ./internal/auth ./internal/app` is 6349 tests green. Policy, Lint, CLI Smoke, Edition, Mock MCP, Windows, cross-platform compile and the race shards are all green in CI.

If you want, the cheap fix is raising `-timeout` on that one job, or splitting `internal/app` out of the macOS shard. Happy to do either here or leave it for a separate PR since it affects main too.

mvanhorn:
The two red Test jobs look like a suite-duration problem rather than a failure in this change.

`internal/app` is hitting Go's default 10-minute limit: the run ends with `panic: test timed out after 10m0s` and `FAIL github.com/DingTalk-Real-AI/dingtalk-workspace-cli/internal/app 600.144s`. The test named as running at that moment had only been going 6s, so it isn't a hang in one test — the package as a whole is just at the edge of the budget. #830 fails the same way.

This PR adds 7 lines to `internal/app/schema_contract_test.go`, which is presumably enough to tip it over on a slower runner.

A re-run would probably be enough to tell whether it's marginal or consistently over, but `gh run rerun` needs admin on the repo so I can't trigger one. If it is consistently over, raising `-timeout` for that package or splitting it would be the durable fix — happy to send that separately if you'd like it.

mvanhorn:
Merged main. This needed more than a conflict fix, because main grew its own @-mention support on `chat message reply` while this sat: `--at-open-dingtalk-ids`, `--at-all`, and automatic `<@id>` normalization, plus the `--group`/`--content` renames.

I resolved additively rather than replacing any of it:

- `--at-users` stays as the userId-accepting flag, matching the `--at-users` that `chat message send-by-webhook` already exposes. It resolves userIds through the existing contact lookup and folds the result into `--at-open-dingtalk-ids`, so both flags share main's single mention path.
- Adopted `--group`/`--content`; the legacy aliases main added keep the old spelling working.
- Declared `at-users` in the reply LeafSpec as `atOpenDingTalkIds`/array, so the generated schema reports the real payload property instead of inferring `atUsers` from the flag name.
- Dropped the seven generated schema JSON artifacts main deleted, and moved the changelog entry into a `.changes/` fragment.

Verified locally: build, the full `internal/helpers` suite, and the schema contract test pass, and these gates are green: release-fragments, skill-commands, generated-drift, param-concepts, param-alias-cooccurrence, command-surface, schema-catalog, schema-assembly, mono-multi-skill-content. I also checked the new tests aren't vacuous by dropping the resolved ids, which fails two of them.

If you would rather `reply` not carry two mention flags, say so and I will close this.

PR #621, opened 2026-07-14
feat(skills): add  multi-skill framework to DWS

typefield:
## Summary

- Reapply the final multi-skill restructuring from [typefield/dingtalk-workspace-cli#1](https://github.com/typefield/dingtalk-workspace-cli/pull/1) and synchronize the PR branch through upstream `main` at `4d05ea4`.
- Keep the progressive reference indexes and focused product references introduced by the source branch.
- Review and repair the skill content itself: CLI paths and flags, real response envelopes, pagination, safety gates, helper scripts, and cross-skill routing.
- Keep current open-source runtime contracts and newly published capabilities instead of replacing them with older source-branch guidance.

## Overwrite and open-source fix audit

| Classification | Decision |
| --- | --- |
| Current `main` implementation fixes | Preserved. AI table `workflow create/update`, Sheet `import create/get`, Contact enterprise onboarding (`org create`, `account create`, `user invite`), shortcuts, release/security work, and current schema contracts are all retained. |
| Upstream fixes that supersede earlier review findings | Preserved after checking the built CLI. `doc export` now genuinely supports `docx`, `markdown`, and `pdf` and requires `--output`; Wiki now exposes `--query` plus `wiki node move/copy`. These are no longer documentation-only claims and must not be overwritten with the older limitations. |
| Legitimate source-branch fixes | Kept. This includes real response compatibility for calendar, attendance, contact, mail, minutes and report; drive recursion via `fileId`; legal 7-character table IDs; current AI table export flags; and the `linkedTableId` validator correction in mono and multi. |
| Structural source-branch overwrite | Kept only where it splits oversized Skill documents into progressive indexes and focused references. Runtime meaning was reconciled with current `main`. |
| Merge conflict resolution | Unioned the capabilities. AI table keeps progressive routing and the new workflow write docs; Sheet keeps progressive routing and the new workbook import flow; Contact keeps roles/dismission/profile plus enterprise onboarding; Doc keeps executable current node/ID mappings. |

## Preserved current-main behavior

- AI table workflow create/update/list/get/enable/disable is documented and dispatchable; only workflow deletion remains unavailable.
- Local xlsx/xls conversion uses `dws sheet import create`; stale `drive upload --convert` guidance is not retained.
- Contact role queries, dismission/profile response handling, and enterprise onboarding coexist.
- Document PDF/Markdown export and Wiki node move/copy are retained because the latest open-source CLI now implements them.
[...]

PR #821, opened 2026-07-29
fix(e2e): move multi-layout account safety assertion to dingtalk-profile

typefield:
## Summary
- Release run 30437390088 failed the MultiSkill e2e because dc20ddec (merged via #621) slimmed `dws-shared/SKILL.md` into a routing entry and moved all account-selection rules into the dedicated `dingtalk-profile` skill, while the e2e contract (2d143589) still asserted the 禁止选择第一项、最近登录或最近使用账号 rule inside `dws-shared/SKILL.md`. This blocked v1.0.55-beta.6.
- Fix direction (option B, reviewed): keep `dws-shared` slim as the refactor intended, and move the multi-layout assertion in `scripts/dev/test-multi-profile-e2e.sh` to `dingtalk-profile/SKILL.md`, which owns the rule in the multi layout. The mono assertion is unchanged.
- New PR-level regression contract `internal/app/multi_skill_safety_contract_test.go`: materializes both embedded layouts and asserts the rule is present (multi: `dingtalk-profile/SKILL.md`, mono: `SKILL.md`), so losing the rule fails at PR time instead of at release time. The new test file also makes this revision full-suite so all quality gates run.

> ⚠️ This adjusts the assertion location of the security contract introduced in 2d143589 — requesting sign-off from @audanye-sudo. Known tradeoff: in the multi layout, agents that never load `dingtalk-profile` (progressive loading) do not see this rule in the shared layer; the rule text itself survives verbatim in `dingtalk-profile/SKILL.md:40` and `mono/SKILL.md:68`.

## Test plan
- [x] `go test ./internal/app/ -run TestMultiSkillAccountSafetyRuleContract` — pass (both layouts)
- [x] `bash scripts/dev/test-multi-profile-e2e.sh` — full chain pass locally with the relocated assertion
- [x] `skills/multi/dws-shared/SKILL.md` and `.github/workflows/ci.yml` — zero diff vs main (final PR diff = 2 files)


haofeng0705 (approved):
LGTM

PR #834, opened 2026-07-30
fix(event): prevent subscription retry storms

wxianfeng:
## Summary

- Add a cross-process retry guard for all 16 public personal-event subscriptions and all five rule types.
- Atomically preclaim multi-event batches; enforce in-flight leases, deterministic backoff/jitter, `Retry-After`, terminal holds, CAS completion, and fail-closed state handling.
- Preserve `retryable` as a three-state contract, expose retry timing/server diagnostics, and add `X-Cli-Version` plus `User-Agent: dws-cli/<version>`.
- Keep one subscription-create HTTP request per logical subscription per CLI invocation. The documented `0/2/1` retry budget is an Agent/host orchestration contract, not a hidden CLI-persisted attempt cap.

This prevents deterministic subscription failures, including cross-organization group rejection, from becoming an unbounded callback retry storm.

## Error and duplicate contract

- The formal `/subscription/user` duplicate path already returns success with the existing `subId`; new and expired-restored subscriptions also return success.
- An API error is never converted to success merely because `details.subscribe_id` is present. There is no unverified duplicate-code allowlist.
- An unknown legacy/proxy `4xx + details.subscribe_id` remains an error, but uses unknown cooldown instead of a one-hour `terminal_hold`.
- Explicit `retryable=false`, stable terminal business codes, and HTTP 401/403 remain non-retryable. HTTP 408/425/429/5xx and network timeouts remain retryable.
- Local run-state I/O failures use unknown cooldown. Cancellation releases the claim, and incomplete reservations fail closed without dereferencing a nil store/claim.

Blast radius: the formal active-duplicate success path is unchanged. Legacy/proxy error responses that old `main` treated as success are now surfaced as errors and cooled down; they are not guessed into success or escalated to a one-hour hold unless the server explicitly marks them terminal.

## Acceptance criteria and executable evidence

| Requirement | Test evidence |
|---|---|
| All 16 public events / five rule types use the guard | `TestCrossPlatformCoveragePersonalSubscriptionProtectionCoversAllPublicEvents` |
| 100 equivalent concurrent attempts produce one callback request | `TestCrossPlatformCoveragePersonalSubscriptionProtectionConcurrentHundredAllowsOneHTTPRequest` |
| Profile, event, target, filter, and endpoint scope fingerprints | `TestCrossPlatformCoveragePersonalSubscriptionFingerprintChangesWithLogicalInputs` |
| Retryability tri-state, Retry-After, terminal codes, unknown duplicate shapes | `TestCrossPlatformCoveragePersonalSubscriptionFailureClassification`, `TestCrossPlatformCoverageClientCreateSubscriptionRejectsErrorsWithSubscribeID` |
| Atomic batch preclaim, partial failure cleanup, crash/CAS recovery | `TestCrossPlatformCoveragePersonalSubscriptionBatchClaimFailureMakesZeroCreateCalls`, `TestCrossPlatformCoverageAttemptStorePartialFailureClearsSuccessAndReleasesUnexecuted`, `TestCrossPlatformCoverageAttemptStoreCASRejectsOldCompletion` |
| Corruption, lock/write failure, permissions, field clamps, and privacy | `TestCrossPlatformCoverageAttemptStoreFailsClosedOnCorruptStateAndWriteFailure`, `TestCrossPlatformCoverageAttemptStoreTightensStateAndLockPermissions`, `TestCrossPlatformCoverageAttemptStoreBoundsPersistedDiagnostics`, `TestCrossPlatformCoverageAttemptStorePersistsNoRawFingerprintInputs` |
| Local I/O cooldown and incomplete reservation guard | `TestCrossPlatformCoveragePersonalSubscriptionSingleAttemptCompletionErrors`, `TestCrossPlatformCoveragePersonalSubscriptionAttemptGuardHelperEdges` |
| CLI headers | `TestCrossPlatformCoverageNewPersonalEventControlClientSetsVersionHeaders`, `TestCrossPlatformCoverageClientBusinessErrorPreservesRetryContractAndClientHeaders` |
| Independent dev-app retry stays bounded | `TestCrossPlatformCoverageDevAppEventSubscribeRunnerFailureIsNotRetried`, `TestCrossPlatformCoverageCallToolDevAppEventSubscribeRetriesAreBounded` |

All 54 changed/new test entry points used for this PR are wired into the repository's `TestCrossPlatformCoverage*` native selector.

## Local guard state

- Default open-edition path: `~/.dws/events/open/personal_stream/<identity_hash>/personal_subscription_attempts.json`. `DWS_CONFIG_DIR` and edition overrides move the config root.
- Identity directory mode: `0700`. JSON and lock files: `0600`.
- Failure streak resets after 24 hours without another failure. `terminal_hold` lasts one hour.
- Normal recovery is to wait for `next_retry_at`. Emergency recovery: first ensure no create process is running for that identity, then delete only `personal_subscription_attempts.json`, not the lock file. This clears all guard records for that identity.

## Verification

- Initial commit `ec996548` had two PR-caused native failures (macOS and Windows) at `37.7224% (843 executable statements)`; aggregate `Coverage` was the derived third red check.
- Follow-up `2808e71c` wires the tests into the enforced selector.
- Follow-up `b82e9754` fixes the Windows-only test cleanup failure exposed by the expanded selector: the audit-runtime coverage test restored a live process-wide sink while incorrectly marking it uninitialized, so a later runner overwrote its owner and leaked the original `.audit.lock` handle. The test now restores the sink as initialized and asserts that a subsequent setup returns the same owner.
- The CI-equivalent command passes after both fixes:

  ```text
  GOWORK=off make coverage-gate-platform BASE_REF=9aa76ea74860168cb62e5becfdbcfbe129608520 PROFILE=/tmp/pr834-coverage-native-final.txt
  changed code coverage: 100.0000% (846 executable statements; target 100.0000%)
  ```

- `GOWORK=off go build ./cmd/...`
- `GOWORK=off go vet ./internal/app ./internal/errors ./internal/event/personal ./internal/transport ./internal/helpers`
- `GOWORK=off go test -race ./internal/app -run '^(TestCrossPlatformCoverageAuditRuntimeCoverage|TestCrossPlatformCoverageRuntimeRunnerRoutingCoverage|TestCrossPlatformCoveragePersonalSubscriptionProtectionConcurrentHundredAllowsOneHTTPRequest)$' -count=1`
- `make format-check`
- `./scripts/policy/check-generated-drift.sh`
- `GOWORK=off go test ./test/unit -run '^(TestCrossPlatformCoverageEventSkill|TestEventSkill)' -count=1`
- GitHub Actions for `b82e9754`: all checks passed (`24 successful`, one expected changed-packages job skipped), including macOS/Windows native coverage, aggregate Coverage, Policy, and all race jobs.
- The repository-wide pre-commit `make` is not claimed green: it stops at pre-existing staticcheck findings outside this diff. The scoped build, vet, format, generated-drift, native coverage, and race commands above are the reported verification.

## Notes

- Does not modify `dws-wukong`.
- Does not add an unreliable group-ownership preflight or an identity-wide global breaker.
- `event status`, `event stop`, `--subscribe-id`, `--dry-run`, and established Stream reconnection behavior remain outside the create-attempt guard.
- Beta/canary rollout remains a separate release step.

### DIFF
--- a/docs/event-subprocess-contract.md
+++ b/docs/event-subprocess-contract.md
@@ -4,7 +4,7 @@ Defines the stable `dws event consume` subprocess contract so an
 orchestrator can determine when the consumer is ready, stop it cleanly,
 and machine-read why it exited.
 
-Scope of this branch: the four **contract** items below. Reconnect
+Scope of this branch: the five **contract** items below. Reconnect
 resilience (keeping the stream alive across a transient upstream drop) is
 tracked separately and intentionally out of scope here.
 
@@ -91,6 +91,42 @@ Ownership-based cleanup:
   subscription is still present (reuse case preserved).
 - T4c (control): `kill -9` leaves subscribe_id lingering (documented risk;
   we only guarantee SIGTERM is clean, we do not fix kill -9 itself).
+
+### 5. Subscription-create retry budget
+
+This policy covers all 16 public personal-event keys and every logical
+subscription in a multi-event command. It applies only before the ready
+marker; reconnecting an established Stream remains a separate mechanism.
+
+- ID resolution, `event consume`, and later `event status/stop` must use the
+  same `--profile`. A user or conversation ID resolved under another profile
+  must not be reused for the current subscription.
+- A logical subscription is keyed by the current profile/identity, event key,
+  rule type, target, and filters. A new `subscribe_id`, `trace_id`, or process
+  does not reset its retry budget.
+- `retryable=false` means `max_additional_attempts=0`.
+- `retryable=true` means `max_additional_attempts=2`. A caller must honor
+  `retry_after_seconds` or `next_retry_at` when present and must not retry
+  early.
+- An omitted retryable value (`retryable=unknown`) means
+  `max_additional_attempts=1`; a second unknown failure stops the operation.
+- `in_flight` means the original logical request is still running.
+  `cooldown` and `terminal_hold` mean a guard is already delaying or blocking
+  it. These states must not recursively launch `event consume`, start a
+  parallel equivalent subscription, or reset the budget with a new subId or
+  trace. The caller waits for the original request/guard or stops.
+- A multi-event command remains one original operation. A caller must not
+  split out a failed event, reorder events, or restart the command to bypass
+  a budget. Existing startup rollback cleans subscriptions created before a
+  later item fails.
+
+**Verification**
+- T5a: non-retryable, retryable, and unknown failures allow respectively
+  0, 2, and 1 additional attempts for the same logical subscription.
+- T5b: a changed subId/trace or process restart does not increase the budget.
+- T5c: `in_flight`/`cooldown` does not recursively issue another create.
+- T5d: multi-event startup cannot be split or reordered to bypass the guard,
+  and a partial startup still rolls back earlier subscriptions.
 
 ## Out of scope (next branch)
 
--- a/docs/event-subprocess-contract.md
+++ b/docs/event-subprocess-contract.md
@@ -92,41 +92,72 @@ Ownership-based cleanup:
 - T4c (control): `kill -9` leaves subscribe_id lingering (documented risk;
   we only guarantee SIGTERM is clean, we do not fix kill -9 itself).
 
-### 5. Subscription-create retry budget
+### 5. Subscription-create retry orchestration and local guard
 
 This policy covers all 16 public personal-event keys and every logical
 subscription in a multi-event command. It applies only before the ready
 marker; reconnecting an established Stream remains a separate mechanism.
 
+- The `0/2/1` limits below are an **Agent/host orchestration contract**, not
+  a CLI-enforced persisted total-attempt cap. Each `dws event consume`
+  process sends at most one subscription-create HTTP request for a logical
+  subscription and performs no in-process automatic retry. The CLI persists
+  only the `in_flight`, `cooldown`, and `terminal_hold` guard states; it does
+  not persist or enforce the Agent/host attempt count across invocations.
 - ID resolution, `event consume`, and later `event status/stop` must use the
   same `--profile`. A user or conversation ID resolved under another profile
   must not be reused for the current subscription.
 - A logical subscription is keyed by the current profile/identity, event key,
   rule type, target, and filters. A new `subscribe_id`, `trace_id`, or process
-  does not reset its retry budget.
-- `retryable=false` means `max_additional_attempts=0`.
-- `retryable=true` means `max_additional_attempts=2`. A caller must honor
-  `retry_after_seconds` or `next_retry_at` when present and must not retry
-  early.
-- An omitted retryable value (`retryable=unknown`) means
-  `max_additional_attempts=1`; a second unknown failure stops the operation.
+  does not create a new logical operation or reset the Agent/host budget.
+- For the Agent/host, `retryable=false` means
+  `max_additional_attempts=0`.
+- For the Agent/host, `retryable=true` means
+  `max_additional_attempts=2`. It must honor `retry_after_seconds` or
+  `next_retry_at` when present and must not retry early.
+- For the Agent/host, an omitted retryable value
+  (`retryable=unknown`) means `max_additional_attempts=1`; a second unknown
+  failure stops the operation.
 - `in_flight` means the original logical request is still running.
   `cooldown` and `terminal_hold` mean a guard is already delaying or blocking
   it. These states must not recursively launch `event consume`, start a
-  parallel equivalent subscription, or reset the budget with a new subId or
-  trace. The caller waits for the original request/guard or stops.
+  parallel equivalent subscription, or bypass the guard with a new subId or
+  trace. The caller waits for the original request/guard or stops, while the
+  Agent/host keeps its own orchestration count.
 - A multi-event command remains one original operation. A caller must not
   split out a failed event, reorder events, or restart the command to bypass
   a budget. Existing startup rollback cleans subscriptions created before a
   later item fails.
 
+#### Local guard state operations
+
+- The default open-edition state file is
+  `~/.dws/events/open/personal_stream/<identity_hash>/personal_subscription_attempts.json`.
+  The config root follows `DWS_CONFIG_DIR` when set, and another edition uses
+  that edition's directory instead of `open`.
+- The identity directory is mode `0700`; both
+  `personal_subscription_attempts.json` and
+  `personal_subscription_attempts.lock` are mode `0600`.
+- A failure streak resets after 24h without another failure. A
+  `terminal_hold` lasts 1h. Prefer waiting until the reported
+  `next_retry_at`; do not clear the file as a normal retry mechanism.
+- For emergency recovery, first ensure that no subscription-create process is
+  running for that identity. Delete only
+  `personal_subscription_attempts.json`, never the lock file. This clears
+  every protection record for that identity, not just one event.
+
 **Verification**
-- T5a: non-retryable, retryable, and unknown failures allow respectively
-  0, 2, and 1 additional attempts for the same logical subscription.
-- T5b: a changed subId/trace or process restart does not increase the budget.
+- T5a (policy): skill/docs tests pin the Agent/host 0/2/1 orchestration
+  contract and explicitly reject describing it as a CLI-persisted hard cap.
+- T5b (CLI): one process issues at most one create request per logical
+  subscription; a changed subId/trace or process restart does not bypass the
+  persisted fingerprint guard.
 - T5c: `in_flight`/`cooldown` does not recursively issue another create.
 - T5d: multi-event startup cannot be split or reordered to bypass the guard,
   and a partial startup still rolls back earlier subscriptions.
+- T5e: state-store tests cover `0700`/`0600` permissions, 24h reset, 1h
+  `terminal_hold`, and identity-scoped cleanup; skill/docs tests pin the
+  operational recovery instructions.
 
 ## Out of scope (next branch)
 
--- a/internal/app/audit_coverage_more_test.go
+++ b/internal/app/audit_coverage_more_test.go
@@ -199,6 +199,14 @@ func TestCrossPlatformCoverageAuditRuntimeCoverage(t *testing.T) {
 		sharedAuditSink = previousSink
 		loadTokenForProfile = previousLoader
 		auditSinkOnce, auditCloseOnce = sync.Once{}, sync.Once{}
+		// The process-wide sink was initialized by TestMain. Preserve that
+		// initialized state when restoring it: leaving auditSinkOnce unused
+		// lets a later runner overwrite the live sink without closing its
+		// .audit.lock handle, which makes TestMain cleanup fail on Windows.
+		auditSinkOnce.Do(func() {})
+		if got := setupAuditSink(); got != previousSink {
+			t.Errorf("restored audit sink = %T, want original %T", got, previousSink)
+		}
 		resetAuditIdentityCache()
 	})
 
--- a/internal/app/coverage_edges_test.go
+++ b/internal/app/coverage_edges_test.go
@@ -1657,12 +1657,13 @@ func TestCrossPlatformCoveragePersonalEventCommandRuntimeCoverage(t *testing.T) 
 		CorpID: "corp", UserID: "user", ClientID: "client",
 	})
 	t.Setenv("DWS_CONFIG_DIR", configDir)
-	var cancelCount int
+	var subscribeCount, cancelCount int
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.URL.Path {
 		case "/event/sublist":
 			_ = json.NewEncoder(w).Encode(map[string]any{"items": []map[string]any{{"subId": "sub", "eventKey": personal.EventMention, "ruleType": "at", "status": "active", "sourceId": "open"}}, "total": 1})
 		case "/subscription/user":
+			subscribeCount++
 			_ = json.NewEncoder(w).Encode(map[string]any{"success": true, "result": []string{"created"}})
 		case "/subscription/cancel":
 			cancelCount++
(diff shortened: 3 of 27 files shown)


haofeng0705:
Review status: **HOLD / request changes**. Two blockers, plus description gaps.

**B1 — the enforced coverage gate fails; the reported one does not run.**
The same 843 changed statements measure differently under the two repo gates:
- full-suite profile → `100.0000% (843 executable statements)` — your number reproduces exactly, verbatim;
- `coverage-gate-platform` (macOS/Windows, the one CI enforces) → **37.7224%, EXIT=2**.

Root cause confirmed locally: the native gate runs only
`-run '^(TestAllShortcuts|TestCrossPlatformCoverage)'`, and the new tests almost never
carry that prefix (`event_personal_attempts_test.go` 17 tests / 0 prefixed;
`attempt_store_test.go` 21 / 0; `internal/event/personal/client_test.go` 24 / 0;
`internal/errors/errors_test.go` 12 / 0; `internal/transport/client_test.go` 18 / 0).
Running *all* tests in the changed packages against the identical gate yields
`100.0000% (843)`, EXIT=0 — so the fix is scoped to test naming, not production code.
The repo already has 1132 `TestCrossPlatformCoverage*` instances, so this is an
established convention. Net effect today: ~1400 new production lines are never
executed in the macOS/Windows coverage jobs. `Coverage` (4s) is a derived failure
of the "Verify coverage profile jobs" guard, not an independent one.

Please also correct the Verification checklist: it checks off
"Changed-code coverage gate: 100.0000%" while three coverage jobs are red, and it
does not disclose them.

**B2 — the duplicate-code allowlist needs server evidence and a stated blast radius.**
`isDuplicateSubscriptionCode` narrows idempotent success from "`Details["subscribe_id"]`
is non-empty" to `DUP` / `DUPLICATE_SUBSCRIPTION` / `SUBSCRIPTION_ALREADY_EXISTS`.
Verified with a throwaway probe on this branch (all four were idempotent successes on
`origin/main`):

    code=DUP                        => IDEMPOTENT SUCCESS subscribe_id=sub-existing-123
    code=SUBSCRIPTION_ALREADY_EXIST => ERROR non_retryable reason=personal_subscription_http_rejected
    code=ALREADY_SUBSCRIBED         => ERROR non_retryable reason=personal_subscription_http_rejected
    code=DUPLICATE                  => ERROR non_retryable reason=personal_subscription_http_rejected

A non-allowlisted duplicate code therefore escalates to a **1-hour `terminal_hold`**.
The only in-repo evidence for the allowlist is `"DUP"` in
`internal/event/personal/coverage_edges_test.go:220`. Could you attach the actual
codes the subscription service returns, and state the blast radius in the PR body?

**Minor**
- `personalSubscriptionLocalFailure(cause error)` — `cause` is unused, and local I/O
  failures (e.g. save run state) are forced to `non_retryable` → the same 1-hour hold,
  while also rolling back a remote subscription that was just created successfully.
- The new contract items T5a–T5d have no executable test. `max_additional_attempts`
  appears in Go only as doc-grep string literals in `test/unit/skill_docs_policy_test.go`,
  and the implementation has no attempt-count cap at all (only 30s/2m/10m/30m backoff),
  so only `retryable=false` is machine-enforced. Either add the cap or soften the doc.
- `completeFailure`'s cancellation path calls `r.store.Release(r.claim)` without the
  nil guard `completeSuccess` has.

**Description gaps (D3/D5/D6/D7)**
- D3: no acceptance criteria and no requirement→test mapping (nothing maps
  "16 events / 5 rule types" to tests).
- D5: no mention of the new state file `personal_subscription_attempts.json`
  (location, 0600/0700 perms, how a user clears a stuck `terminal_hold`), the
  `attemptFailureReset=24h` / `attemptTerminalHold=1h` constants, or the new
  `X-Cli-Version` / `User-Agent` headers.
- D6: exact commands are missing. For the coverage claim the command is decisive —
  please state the invocation and base-ref.
- D7: the two PR-caused native coverage failures are not disclosed. Also, the claimed
  "unrelated local/baseline blockers (macOS codesign, stdio timeout, commit-hook
  fixtures)" could not be verified — my local sweep of 40 packages incl. all changed
  packages was 0 FAIL; please give the exact command that reproduces them.

**Reproduced clean:** `go build ./cmd/...`, targeted `go vet`, `make format-check`,
`check-generated-drift.sh`, and
`TestPersonalSubscriptionProtectionConcurrentHundredAllowsOneHTTPRequest` under `-race`
(0.87s, no race reports). Security posture of the new store is solid: 0700 dir / 0600
file, atomic tmp→chmod→rename, 256-byte field clamps on both write and load paths,
24h pruning, and identity-scoped fingerprints (no cross-user collision).
The concurrency design itself I have no objection to.

wxianfeng:
@haofeng0705 已逐项复核，接受 **HOLD**，当前不应合并。

**B1 成立。** 使用固定 base `9aa76ea74860168cb62e5becfdbcfbe129608520`：

`GOWORK=off make coverage-gate-platform BASE_REF=9aa76ea74860168cb62e5becfdbcfbe129608520 PROFILE=/private/tmp/pr834-coverage-native.txt`

可复现 `37.7224% (843 executable statements)`, exit 2。macOS、Windows 两个 native job 失败，`Coverage` 是派生失败。根因确实是新增测试没有进入 `TestCrossPlatformCoverage*` selector；同三包运行全部测试为 100%，因此 B1 只需修正 coverage-driver 测试命名/接线，不改生产逻辑。Verification checklist 会同步纠正并披露三个红项。

**B2 成立。** 当前 `/subscription/user` 的正式行为是：重复 active 订阅直接 success 并返回 existing subId，expired 恢复和新建也返回 success；校验错误包括 `ILLEGAL_PARAMETER`、`SOURCE_ID_REQUIRED`、`UNIFIED_APP_ID_NOT_FOUND`、`EVENT_KEY_NOT_SUPPORTED`，其他异常统一为 `SYSTEM_ERROR`；portal 另有 `NO_AUTH`、`INVALID_PARAM`。没有证据表明当前链路会返回 `DUP`、`DUPLICATE_SUBSCRIPTION`、`SUBSCRIPTION_ALREADY_EXISTS`。

相比 main，当前分支会把“非 allowlist 错误但携带 `details.subscribe_id`”从幂等成功改成失败；当前正式重复订阅路径不受影响，但 legacy/proxy 的 4xx 形态可能进入 1 小时 `terminal_hold`。后续不会猜测扩充错误码：未知 code 仍不得误判成功，但应按 unknown cooldown 处理，只有服务端明确 `retryable=false` 或稳定终态业务码才进入 terminal hold。

**Minor 处理结论：**

- 本地 run-state I/O 错误固定为 non-retryable、回滚远端后仍 hold 1 小时，问题成立；应改为 unknown/cooldown，并清理未使用的 `cause`。
- `completeFailure` 缺少不完整 reservation 的防御性检查，问题成立。
- `0/2/1` 是 Agent orchestration contract，不是 CLI 持久化总次数上限；文档会明确这一点。CLI 的单进程零重试、退避、并发预占和指纹稳定性继续由行为测试保证，本分支不额外引入隐藏的硬次数上限。

PR 描述还会补齐：16 events / 5 rule types → 测试映射、状态文件位置与 0700/0600 权限、24h reset/1h hold、紧急清理方法、CLI headers、精确命令/base-ref，以及三个 coverage 失败。无法给出复现命令的 codesign/stdio/commit-hook blocker 声明将删除。

完成修复并让 native gate 达到 100% 后再请求 re-review。

wxianfeng:
@haofeng0705 CR 阻塞项已修复，提交 `2808e71c`，请 re-review。

**B1 已修复。** 所有本 PR 新增/变更的 54 个测试入口均已接入 `TestCrossPlatformCoverage*` selector。使用原 review 的固定 base：

```text
GOWORK=off make coverage-gate-platform BASE_REF=9aa76ea74860168cb62e5becfdbcfbe129608520 PROFILE=/tmp/pr834-coverage-native-final.txt
changed code coverage: 100.0000% (846 executable statements; target 100.0000%)
```

初始 `ec996548` 的 macOS、Windows 两个 native 失败和派生 `Coverage` 红项已在 PR 描述中如实披露。

**B2 已修复。** 已删除未经服务端契约证明的 duplicate allowlist；任何 API error 即使带 `details.subscribe_id` 也保持 error。正式 active duplicate 的 `success + existing subId` 路径不变。未知 legacy/proxy `4xx + subId` 走 unknown cooldown，不再误升为一小时 hold；显式 `retryable=false`、稳定终态码和 401/403 仍为 non-retryable。

**Minor 已处理。**

- 本地 run-state I/O 错误改为 unknown/cooldown，并移除未使用的 `cause` 参数。
- `completeFailure` 对不完整 reservation 增加 fail-closed guard，取消路径不再可能解引用 nil store/claim。
- `0/2/1` 已明确为 Agent/host 编排约束；CLI 保持单进程零自动重试，不宣称持久化硬次数上限。

PR 描述已补齐 16 events / 5 rule types 的 requirement→test 映射、状态文件路径与权限、24h reset/1h hold、紧急清理、CLI headers、精确命令/base-ref 和 duplicate blast radius。

本地验证：native gate 100%、100 并发 race、build、targeted vet、format-check、generated drift、Skill policy 通过。隔离本机 Git hook 后的全量运行只出现 4 个非稳定、非本分支夹具失败（macOS codesign、2 个 Gitee reconciliation、1 个 release preflight timeout），同环境逐一 `-count=1` 复跑全部通过，修改包无稳定失败。

wxianfeng:
@haofeng0705 追加修复已推送：`b82e9754`。本轮 CR 的 B1、B2 和已确认 Minor 均已处理，现请 re-review。

**Windows native follow-up 已修复。** 扩大 selector 后暴露的是测试生命周期问题，不是订阅生产逻辑：`TestCrossPlatformCoverageAuditRuntimeCoverage` 恢复了 TestMain 的进程级 audit sink，却把 `auditSinkOnce` 留成“未初始化”；后续 runner 新建 sink 并覆盖旧 owner，原 `.audit.lock` 句柄失去引用。Unix 允许删除仍打开的文件而掩盖了问题，Windows 在 TestMain `RemoveAll` 时失败。

修复保持原 sink 为 initialized，并增加 identity invariant：恢复后再次 `setupAuditSink()` 必须返回同一个 owner，确保 TestMain 能关闭原锁句柄。没有修改 `internal/audit` 的锁实现，也没有修改订阅生产路径。

验证结果：

```text
GOWORK=off make coverage-gate-platform BASE_REF=9aa76ea74860168cb62e5becfdbcfbe129608520 PROFILE=/tmp/pr834-coverage-native-lock-fix.txt
changed code coverage: 100.0000% (846 executable statements; target 100.0000%)
```

- audit 生命周期 + runner + 100 并发订阅保护的 race 回归通过；
- build、targeted vet、format-check、generated drift 通过；
- GitHub Actions：`24 successful`、`1 skipped`（预期 changed-packages job），所有 required checks 已通过；
- macOS/Windows native coverage、aggregate Coverage、Policy、Windows test 和全部 race jobs 均为绿色。

同时更正上一条评论的 verification 表述：PR 描述已删除 codesign/Gitee/preflight 等非本 PR blocker 声明，不再将其作为本 PR 验证结论；仓库级 pre-commit `make` 仍会停在本 diff 外的既有 staticcheck 项，PR 只报告上面实际通过的 scoped commands。

haofeng0705:
Re-review of `b82e9754`. Both blockers verified fixed, all three minors fixed, and the description gaps are closed. **No further blockers from me.**

### B1 — verified fixed against the gate CI actually enforces

I re-ran the same command shape as round 1, on the same base-ref, so the two numbers are directly comparable:

```text
./scripts/policy/run-platform-coverage-gate.sh --base-ref 9aa76ea74860168cb62e5becfdbcfbe129608520
round 1 (ec996548): changed code coverage:  37.7224% (843 executable statements)   EXIT=2
round 2 (b82e9754): changed code coverage: 100.0000% (846 executable statements)   GATEEXIT=0
```

Your figure reproduces verbatim, under the native selector — not the full-suite profile. The `54` claim also checks out exactly: `git diff ec99654..HEAD | grep -cE "^\+func (TestCrossPlatformCoverage|TestAllShortcuts)"` → `54`. No production code was moved to make the number, which was the point.

### B2 — verified fixed, and the ordering is right

`isDuplicateSubscriptionCode` and the whole allowlist are gone; no `APIError` converts to success anywhere in `createSubscription`. Reading the switch in `classifyPersonalSubscriptionFailure`, the new
`personalSubscriptionErrorHasSubscribeID` case sits **after** 401/403 and after terminal business codes, and leaves `retryability` at its initialized `Unknown` → cooldown rather than the 1-hour hold. Precedence is therefore not weakened, and your own subtests pin exactly that:
`403_with_existing_id_remains_an_auth_rejection`, `legacy_DUP_with_existing_id_remains_an_unknown_error`, `legacy_{SUBSCRIPTION_ALREADY_EXIST,ALREADY_SUBSCRIBED}_...`.

The blast-radius paragraph in the body is accurate rather than softened: old `main` converted to success exactly when `Details["subscribe_id"]` was non-empty, which is precisely the set now routed to cooldown. Same set, no hand-waving.

### Minors

- nil guard on the incomplete-reservation path — present, and it fails closed via `personalSubscriptionGuardError` instead of dereferencing.
- `personalSubscriptionLocalFailure()` — unused `cause` removed, downgraded to `Unknown`; both call sites updated.
- `max_additional_attempts` `0/2/1` — reframed as an Agent/host orchestration budget, and this one is now **machine-enforced**, not just reworded: `TestCrossPlatformCoverageEventSkillPinsSubscriptionRetryOrchestrationContract` requires the docs to literally carry `does not persist or enforce the Agent/host attempt count` (and the zh equivalent), failing with *"overstates CLI retry enforcement"*. That is a stronger fix than I asked for — it cannot silently rot back.

### `b82e9754` matches its description

1 file, 8 insertions, test-only. `internal/audit` locking untouched; subscription production path untouched. The fix marks the restored process-wide sink `initialized` and adds the owner-identity invariant, so a later runner cannot overwrite the live sink and orphan its `.audit.lock` handle. That is a genuine test-lifecycle bug that Unix hid by allowing unlink-while-open — the diagnosis is right, and the invariant assert is the part that keeps it from regressing.

### Description D1–D7

All seven now satisfied; the four I flagged are closed. All **18** test names in the acceptance table exist in the tree (checked each one), so D3 is a real requirement→test mapping rather than aspirational. D7 now discloses the two original native failures with the 37.7224%/843 figures, and correctly retracts the earlier unverifiable blocker claims.

I confirmed the remaining staticcheck statement is a **baseline** finding, not yours: repo-wide 19 findings, exactly 1 in a file this PR touches — `internal/app/coverage_edges_test.go:357` (SA1012) — and `git blame` puts that line on `8633246e`, which is an ancestor of the base. Your wording ("stops at pre-existing staticcheck findings outside this diff") is accurate.

### My own checks on `b82e9754`

`go build ./cmd/...` OK · targeted `go vet` OK · `gofmt -l` clean · `check-generated-drift.sh` ok · `-race` on the audit-lifecycle + 100-concurrent-subscription + classification + guard-helper tests: `ok internal/app 4.738s`, 0 data races.

### One informational note, not a blocker

A generic `4xx` carrying neither a terminal business code nor `details.subscribe_id` still falls through to `>= 400` → `non_retryable` → 1-hour `terminal_hold`. That is new relative to `main` (which had no guard at all), and it is the intended fail-closed posture — I raise it only because the 1-hour window is the most user-visible edge of this PR. It is documented, including the emergency-recovery procedure and its identity-wide blast radius, so I am not asking for a change. Worth watching in canary: if a benign, transient-but-non-5xx shape shows up in the wild, it will cost an hour.

Gate result on my side: pass. Recorded as a comment — I have not filed a formal approving review, so this does not by itself satisfy the repo's review requirement. Note also that the green CI is not what I based this on; the decisive evidence is the locally reproduced native gate above.

haofeng0705 (approved):
LGTM

PR #923, opened 2026-08-10
feat(dev): integrate unified results into dingtalk-dev

typefield:
## Status

Draft follow-up to #924. Do not merge before the framework-core PR lands. After #924 is merged, this branch will be rebased/refreshed so the diff contains only product integration.

## Follow-up scope

- migrate selected `dev ...` and `devapp +...` terminal commands per command
- add dingtalk-dev result projection for success, pending, partial failure, and pagination
- preserve both existing command prefixes and keep `--format json` as the only Agent-facing output selector
- keep foreground streaming commands on legacy output until a stream contract exists

## Explicit exclusions

- no public `--output-contract`
- no `contract_version` response field
- no command-prefix rename or redirect
- no checked-in JSON fixtures

## Dependency

- #924 framework core


typefield:
Superseded by #924. Per the latest scope decision, the unified framework and dingtalk-dev/devapp pilot now ship in one reviewable PR; this older full-pilot branch also contains overlapping and unrelated changes, so it should not be merged.

PR #918, opened 2026-08-08
fix(devapp): preserve pagination across list shortcuts

PeterGuy326:
## Summary

- preserve `hasMore` and `nextCursor` in every paginated `devapp` shortcut output: `+list`, `+permission-list`, `+event-list`, and `+version-list`
- forward opaque request cursors byte-for-byte and require an advancing non-empty cursor whenever `hasMore=true`
- normalize deployed terminal response variants (missing, `null`, empty, or last-observed cursor) to `nextCursor: ""`
- fail closed when list and pagination metadata are malformed, ambiguous, split across envelopes, or otherwise unsafe to continue
- return contract-correct terminal envelopes for all four paginated tools under `--mock`

## Root cause

The MCP providers already return pagination metadata, but the CLI shortcuts projected their responses into only the item collection plus `count`. This discarded the continuation contract and made a non-terminal page indistinguishable from a complete result.

The original revision covered only `devapp +list`. Source and live MCP inspection found the same projection loss in `+permission-list`, `+event-list`, and `+version-list`, plus two deployed terminal shapes: `list_dev_app` may retain the last observed cursor with `hasMore=false`, while `list_dev_app_versions` may omit a null terminal cursor entirely.

## Contract details

- `hasMore=true` requires a non-empty string cursor different from the request cursor.
- `hasMore=false` is authoritative; a missing, null, empty, or last-observed provider cursor is normalized to an empty public cursor.
- `--cursor` uses the raw pflag value. Leading/trailing whitespace and whitespace-only values are forwarded byte-for-byte rather than trimmed. This is intentional for an opaque token and keeps stalled-cursor comparison on the exact value sent to the provider.
- Per DEC-DWS-917-R1B, an unprojectable app item is a page-integrity failure: silently dropping it could produce a false terminal “not found” result. It therefore retains the frozen `devapp_pagination_contract_invalid` page-contract reason.
- The scope extension is intentional and limited to the four `devapp` shortcuts that share the same cursor/page-size contract and were confirmed to discard provider pagination metadata. Other devapp tools remain passthrough behavior.

## Mock compatibility

Mock mode now returns each tool's real collection key with a terminal envelope:

- `list_dev_app` → `apps`
- `list_dev_app_permissions` → `items`
- `list_dev_app_events` → `events`
- `list_dev_app_versions` → `items`

End-to-end tests execute all four public shortcuts with `--mock` and require exit 0 plus exactly the projected collection, `count`, `hasMore=false`, and `nextCursor=""`.

## User impact

[...]

PR #929, opened 2026-08-10
fix(devapp): preserve pagination across list shortcuts

typefield:
## Summary

- preserve `hasMore` and `nextCursor` in all paginated `devapp` shortcut outputs: `+list`, `+permission-list`, `+event-list`, and `+version-list`
- forward opaque request cursors unchanged and require an advancing non-empty cursor whenever `hasMore=true`
- normalize deployed terminal response variants (missing, `null`, empty, or last-observed cursor) to `nextCursor: ""`
- fail closed when list and pagination metadata are malformed, ambiguous, split across envelopes, or otherwise unsafe to continue

## Root cause

The MCP providers already return pagination metadata, but the CLI shortcuts projected their responses into only the item collection plus `count`. This discarded the continuation contract and made a non-terminal page indistinguishable from a complete result.

Draft PR #918 addressed only `devapp +list`. Live MCP verification also showed its terminal rule was too strict: `list_dev_app` may return the last observed cursor with `hasMore=false`, while `list_dev_app_versions` may omit a null terminal cursor entirely. This PR keeps `hasMore` authoritative on terminal pages while retaining strict validation for non-terminal pages.

## User impact

Each paginated devapp list command now returns exactly its projected collection, `count`, `hasMore`, and `nextCursor`. Existing item projections remain unchanged. Callers can safely continue with `--cursor` and can reliably detect completion.

Refs #917
Supersedes #918

## Verification

- `DWS_PACKAGE_VERSION=0.0.0-test go test -count=1 ./internal/shortcut/devapp`
- `DWS_PACKAGE_VERSION=0.0.0-test go test -race -count=1 ./internal/shortcut/devapp`
- focused `internal/app` devapp and shortcut-schema tests
- `make build`
- generated drift, schema catalog, open-source audit, and CHANGELOG PR policy checks
- live MCP verification:
  - `+list`: non-terminal cursor returned; cursor continuation succeeded
  - `+permission-list`: non-terminal cursor returned; cursor continuation succeeded
  - `+event-list`: non-terminal cursor returned; cursor continuation succeeded
  - `+version-list`: empty terminal page normalized to `hasMore=false, nextCursor=""`


typefield:
Closed as a duplicate: the follow-up fix was appended directly to #918 as commit e023be14, and #918 has been updated to cover all paginated devapp list shortcuts.

PR #1013, opened 2026-08-14
fix(chat): split references and harden card updates

dxb121:
## Summary

- split the chat message and group references into task-focused query, action, media, discovery, and administration guides, then update intent routing to load the smallest relevant reference
- distinguish an accepted streaming-card update request from an independently verified visible update while preserving the real `bizId` across accepted, failed, and unverified outcomes
- explain that `chat +search-msg --from` is ambiguous and direct callers to `--sender` or `--start` instead of guessing intent
- raise the reviewed chat Skill context-budget allowance from 5% to 10% for the reference split and add regression coverage plus a release fragment

## Why

The previous chat references mixed unrelated query, write, media, and group-administration workflows, so an Agent had to load more context than a single task required. The card update path also treated some acknowledgement-shaped responses as stronger evidence than the backend actually supplied, which could encourage a caller to report a visible update without verification or retry an accepted write and duplicate it. Finally, `--from` can reasonably mean either a sender or a time boundary in a message query and should fail with actionable guidance rather than be interpreted silently.

## Risk tier

- [ ] Documentation-only: prose/assets only
- [ ] Standard
- [x] High-risk: this PR changes a Skill context-budget policy threshold and fail-closed card-update evidence semantics

## Compatibility and governance

- Existing public command paths and flags remain available; no flag is renamed, hidden, removed, made required, or changed in type/scope/shorthand.
- `--from` remains unsupported for `chat +search-msg`; only its structured validation guidance is made explicit.
- Interface Integrity reports `compatible: true` with `blocking: []` against both current `main` and stable `v1.0.58`.
- Complete Schema compatibility passes against both references.
- Because there is no public parameter lifecycle or requiredness migration, the protected two-PR migration ledger is not required.

## Verification

- [x] Release fragment: `.changes/chat-reference-card-verification.md`
- [x] `git diff --check`
- [x] `make build`
- [x] affected-package tests: `./internal/app ./internal/shortcut/chat ./internal/shortcut/chatmsg ./internal/shortcut/smart ./internal/shortcut/targetresolver`
- [x] focused card acknowledgement/verification and `--from` ambiguity regressions
- [x] `make policy` — Skill context budget, multi-IM intent chain, live command audit, command surface, generated drift, parameter concepts, aliases, Schema catalog, runtime confirmation truth, and 1,366 Agent examples all pass
- [x] authoritative Interface Integrity — compatible, blocking 0
- [x] complete Schema compatibility — PASS
[...]

PR #998, opened 2026-08-13
fix(chat): harden openDingTalkId target routing

dxb121:
## Summary

- add deterministic user-target resolution across names, `userId`, and current-format `openDingTalkId` values
- validate explicit `openDingTalkId` inputs before execution instead of allowing malformed or ambiguous values to enter fuzzy resolution
- extend `chat +chat-messages` with compatible sender targeting through `--sender` / `--sender-query`, including resolved-filter and sender-scope evidence
- harden `chat +search-msg` sender and conversation routing, fail closed when sender scope cannot be proven, and deduplicate by stable message ID
- align helper and shortcut call paths so IDs are passed to the correct backend fields without changing existing command paths
- add deterministic unit, cross-platform, parameter-alias, compatibility, and Mock MCP coverage
- add the required release fragment for the user-visible routing behavior

## Why

Values in the current `openDingTalkId` format can overlap with ordinary text and legacy identifiers. Routing them through fuzzy name lookup can select the wrong identity or reject a valid explicit target. Message search also needs to prove that returned rows stay within the resolved sender scope; otherwise a successful response can contain out-of-scope messages.

This change gives explicit IDs a strict validation path, keeps natural-language sender lookup available through the query flags, and fails closed when identity or result scope cannot be verified.

## Risk tier

- [ ] Documentation-only: prose/assets only
- [x] Standard: scoped chat identity-routing and read-path contract changes with a stable package graph
- [ ] High-risk: workflow/policy, package graph, generated Schema/registry, platform, auth/keychain, installer, packaging, release, transport, or recovery

## Compatibility and governance

- Existing command paths and existing flags remain available.
- `chat +chat-messages --sender` is an optional compatible flag addition.
- Interface Integrity reports `compatible: true` and `blocking: []` against both the latest `main` merge-base and stable `v1.0.58`.
- Schema compatibility passes against both references.
- This is not a visible-to-hidden rename, requiredness migration, or alias lifecycle change, so the flag-migration governance ledger introduced by #921 is not required.
- No protected workflow, policy, packaging, installer, or release implementation is modified.

## Verification

- [x] Release fragment added: `.changes/open-dingtalk-id-format-routing.md`
- [x] Release-seal validation: N/A; this is an ordinary feature/fix PR, not a release-seal PR
- [x] `make build`
- [x] targeted affected-package tests:
[...]

PR #1029, opened 2026-08-17
fix(chat): preserve unverified sender identity semantics

dxb121:
## Summary

- preserve `unverified_user_id` for mixed `--sender` / `--senders` inputs even when returned messages contain an exact matching `senderId`
- continue delivering exact positive matches while keeping `complete=false`, `identity_unverified`, and negative-conclusion guards until the directory has verified the identity
- align `--sender-query` Skill guidance with the fail-closed Runtime behavior: resolution failures suppress unfiltered messages and return `sender_resolution_failed`
- add focused regressions and the required release fragment

## Why

Follow-up to #998 and #1013.

When directory lookup cannot classify a mixed sender input, Runtime temporarily treats the original value as an unverified userId candidate. The previous post-filter check removed that uncertainty whenever any message happened to carry the same `senderId`. A name that collided with another user's userId could therefore be reported as a verified target and incorrectly allow complete conclusions.

The split Chat Skill references also retained the older statement that `--sender-query` resolution failure preserves all messages, while the merged Runtime correctly fails closed and suppresses unfiltered output.

## Risk tier

- [ ] Documentation-only
- [x] Standard: scoped Chat read/search identity-contract correction with a stable package graph
- [ ] High-risk

## Compatibility and governance

- Existing command paths, flags, aliases, types, requiredness, and parameter scopes are unchanged.
- Exact positive `senderId` matches remain available; only their verification/completeness envelope stays conservative when the input identity is unverified.
- `--sender-query` Runtime behavior is unchanged; its Contract and Skill guidance now describe the existing fail-closed behavior.
- No flag lifecycle or protected parameter migration is involved, so a separate governance PR is not required.

## Verification

- [x] Release fragment: `.changes/chat-sender-identity-contract.md`
- [x] `git diff --check origin/main...HEAD`
- [x] `make build`
- [x] `go vet ./internal/shortcut/smart`
- [x] focused sender-identity regressions:
  `DWS_PACKAGE_VERSION=0.0.0-test go test -count=1 ./internal/shortcut/smart -run '^(TestCrossPlatformCoverageChatMessagesSenderFilterFailureEdges|TestCrossPlatformCoverageSearchMsgStableUserIDContinuesWhenDirectoryIsUnavailable|TestCrossPlatformCoverageSearchMsgResolutionFailureEdges)$'`
- [x] affected package: `DWS_PACKAGE_VERSION=0.0.0-test go test -count=1 ./internal/shortcut/smart`
- [x] native changed-code coverage: **100.0000% (8 executable statements)**
- [x] `./scripts/policy/check-release-fragments.sh origin/main HEAD`
- [x] `./scripts/policy/check-schema-catalog.sh`
- [x] open-source assets, Skill context budget, and multi-IM Skill chain checks
- [x] PII/credential review: changed tests and documentation use synthetic values only; no screenshots or live DingTalk output are included

## Behavior evidence

- `TestCrossPlatformCoverageChatMessagesSenderFilterFailureEdges` proves an exact positive match remains deliverable but retains `identity_unverified`, `complete=false`, and `negativeConclusionAllowed=false`.
- `TestCrossPlatformCoverageSearchMsgStableUserIDContinuesWhenDirectoryIsUnavailable` proves search preserves positive matches while publishing `targetsResolved=false` and an identity-verification failure.
- Skill references now distinguish the mixed `--sender` fallback from name-only `--sender-query` fail-closed behavior.

## Notes

- This PR is intentionally opened as **Draft**. Focused screenshots and any required Agent evidence will be attached before marking it Ready for review.
- The branch was rebased onto current `main`; stable patch-id was unchanged and range-diff showed an equivalent single commit.
- The change is contained in one commit and can be rolled back by reverting it.


dxy704330469 (approved):
CR 通过。评审 SHA：fc05976d3328b19c983984cb379c3e5c1f913590。未发现 P0/P1/P2 可操作问题；Runtime、Contract/Skill 语义及回归测试一致，本地构建、全量测试、Schema/漂移门禁与 changed-code coverage 均通过。

PR #1136, opened 2026-08-25
feat(skills): optimize Report and Sheet workflows

pengzhihan47-star:
## Summary

- Compress Report and Sheet guidance while preserving success-critical contracts and staged reference loading.
- Fix Report latest pagination, duplicate reconciliation, structured error handling, bounded output, readable timestamps, and the midnight query boundary.
- Reuse and return the probed Sheet `sheetId` from `create-with-data` without an extra remote call; preserve both top-level IDs and legacy `.result.nodeId` plus outer `requestId`.
- Keep Sheet write verification aligned with `main` (one readback per non-empty target) and expose structured partial/unknown state after a non-atomic create/write/style failure so callers do not duplicate workbooks.
- Clarify AITable versus Sheet routing, repair GFM table rendering, and reduce unnecessary Help and Catalog discovery.

## Risk tier

- [ ] Documentation-only: prose/assets only; no executable, generated, workflow,
  packaging, or interface behavior changed
- [x] Standard: ordinary implementation change with a stable package graph
- [ ] High-risk: workflow/policy, package graph, generated Schema/registry,
  platform, auth/keychain, installer, packaging, release, transport, recovery,
  or another fail-closed infrastructure change

## Scope

- Report, Sheet, and the required AITable/Sheet routing boundary.
- The 11 `skills/mono` files are Sheet reference mirrors required by repository consistency; no `dingtalk-doc` product files are changed.
- `test/scripts/release_script_test.go` contains the pre-existing release-test flake hardening already present in this PR.

## Verification

- [x] Release fragment added: `.changes/1136-report-sheet-workflow-correctness.md`; release validation passes.
- [x] Release-seal validation: N/A — this is an ordinary PR with a fragment and does not edit or seal `CHANGELOG.md`.
- [x] Targeted test/check commands and results:
  - `make build` — PASS.
  - `go test ./internal/helpers -count=1` — PASS (`ok`, 100.612s).
  - `go test ./test/unit -count=1` — PASS.
  - `go test ./internal/shortcut/report -count=1` — PASS on the preceding Sheet-only delta base.
  - `go test ./internal/app -run TestReviewedRoutedInterfacesReachFinalSchema -count=1` — PASS.
  - `python3 scripts/dev/report_received_today_test.py` — PASS, 7 tests on the preceding Sheet-only delta base.
  - `python3 scripts/gen_skill_shortcut_sections.py --check` — PASS (`visible_shortcuts=401 services=20`).
  - `bash scripts/policy/check-command-surface.sh --strict` — PASS.
  - `bash scripts/policy/check-generated-drift.sh` — PASS, including identical Schema registry hashes.
  - `git diff --check` — PASS.
- [x] Behavior evidence:
  - `TestCrossPlatformCoverageReportLatestRequiresCompleteOrderedListAndExactReadback` locks bounded multi-page latest selection and exact readback.
  - `TestCrossPlatformCoverageAddCreatedSheetIDPreservesLegacyEnvelopeAndProjectsTopLevel` locks top-level `nodeId` / `sheetId` while retaining `.result.nodeId`, outer `requestId`, and exact large integers.
  - `TestCrossPlatformCoverageSheetCreateReadbackMatchesMainSingleAttempt` locks one write readback with no 250/500/1000ms retry loop.
  - `TestCrossPlatformCoverageSheetCreateWriteFailureIsNotBlindlyRetryable` and `TestSheetCreateStyleApplicationFailureReportsIndex` lock structured recovery IDs, stages, status, and `retryable=false` after creation.
  - `test_query_window_clamps_first_second_after_midnight` locks a strictly increasing `--days 1` window at midnight; `test_format_create_time_uses_shanghai_timezone` locks readable `+0800` timestamps.
- [x] Documentation rendering checked: every AITable View/Form/Dashboard/Chart table row now has exactly two GFM columns; Sheet sub-references match their mono mirrors apart from the repository-required relative-link paths.
- [x] Full local suite: N/A for the selected Standard tier; affected packages and contract tests are listed above, while CI runs the complete matrix.
- [x] Package-manager verification: N/A — no packaging or installer surface changed.

## Notes

- `create-with-data` is a multi-request workflow, not a backend transaction. After creation, errors include `details.nodeId`, `details.sheetId` when known, `details.status`, and the failed stage; the whole create command is explicitly non-retryable. Recovery reads the existing workbook and only completes missing steps. It never auto-deletes written data.
- The write-after-readback retry introduced by this PR was removed. Behavior now matches `main`: one read per non-empty target, while retaining probe `sheetId` reuse.
- The received-report helper intentionally fails closed after 10 pages / 200 reports instead of presenting a partial list as complete. High-volume callers should use narrower time windows or the direct paginated Report command.
- Compression retains the restored success-critical limits: Report `--size <= 20`, Report contents <= 10MB, the Sheet chart title/legend/axis field set, and `create-with-data --styles` unknown-key rejection.


haofeng0705 (changes requested):
## Gate decision

REQUEST CHANGES — the Scope statement is contradicted by the diff, no release fragment accompanies several user-visible behavior changes, and three reproduced defects (GFM table truncation, raw epoch timestamps against this PR's own injected contract, a midnight `--days 1` hard failure) remain open.

The engineering here is strong: the Report pagination rewrite is well-bounded and genuinely regression-locked, and every command listed under Tests reproduces cleanly. The blockers below are specific and mostly small.

## Review provenance

- [PASS] Reviewed at PR head `76fd76d08bc11f9650d703ed3dacc61e38dcec7a` in a clean dedicated detached checkout
- [PASS] Base revision / merge-base `76618a6b8c73dac98935da047d65dad5d892e72f`
- [PASS] Head re-checked immediately before submission — same SHA, no drift

## PR description vs repository template

Template graded against `.github/pull_request_template.md`.

- [PASS] T1 Summary — four bullets state what changed; the "why" is implicit but recoverable
- [FAIL] T2 Risk tier — no tier is selected. The template's `## Risk tier` section is absent; the description substitutes a `## Scope` section instead. Given the diff touches a generator (`scripts/gen_skill_shortcut_sections.py`) and generated skill artifacts, state the tier explicitly.
- [FAIL] T3 Release fragment — neither a `.changes/<name>.md` nor an `N/A` with a reason. See the fragment section below.
- [FAIL] T4 Release-seal validation — no `N/A` declaration. (Distinct from T3: this is the missing declaration, not the missing fragment.)
- [FAIL] T5 Targeted test/check commands and results — seven commands are listed, but no result is stated for any of them. I reproduced all seven and they pass, so the commands are right; the template asks for "commands **and results**".
- [FAIL] T6 Behavior evidence — no test name, CLI output shape, or before/after result. The PR does add strong tests (e.g. `TestCrossPlatformCoverageReportLatestRequiresCompleteOrderedListAndExactReadback`); naming one or two would satisfy this.
- [FAIL] T7 Conditional checks — `check-generated-drift.sh` and `check-command-surface.sh --strict` both apply here (generator inputs changed; a contract `Result` spec was added) and neither is declared. Both actually pass — see below — so this is a declaration gap, not a quality gap.
- [PASS] T8 Notes — the `## Note` bullet records the signal-9 termination of the local Schema Catalog generator and the narrower test used instead

**Materially false scope claim (blocking).** `## Scope` states:

> - no dingtalk-doc or skills/mono files

The `dingtalk-doc` half is accurate. The `skills/mono` half is not — this PR changes **11** files under `skills/mono/`, with 499 insertions and **3453 deletions**, i.e. 45% of the PR's total 7702 deletions:

```
skills/mono/references/products/sheet.md
skills/mono/references/products/sheet/sheet-{batch-operations,chart,conditional-format,
  dimension-operations,filter-view,read-data,revision-changeset,style-format,
  workbook,write-data}.md
```

Commit `57ecb1778` ("fix(skills): sync mono Sheet references") added them after the description was written. Please update the Scope bullet — a reviewer trusting it would skip 45% of the deletions.

Observations outside the template — explicitly **not** counted as defects: local OS/Go version, a requirement-to-test mapping table, a limitations section, reproducing the full CI suite locally, and upstream MCP contract verifiability.

## Author validation reproduction

| ID | Acceptance criterion | Author claim / exact step | Documented environment | Independent result | Match | Evidence |
|---|---|---|---|---|---|---|
| V1 | Build succeeds | `make build` | none stated | PASS | Yes | exit 0 |
| V2 | Sheet helpers green | `go test ./internal/helpers -count=1` | none stated | PASS | Yes | `ok ... 102.950s` |
| V3 | Report shortcuts green | `go test ./internal/shortcut/report -count=1` | none stated | PASS | Yes | `ok ... 0.517s` |
| V4 | Routed interfaces reach final Schema | `go test ./internal/app -run TestReviewedRoutedInterfacesReachFinalSchema -count=1` | none stated | PASS | Yes | `ok ... 1.079s` |
| V5 | Python suite green | `python3 scripts/dev/report_received_today_test.py` | none stated | PASS | Yes | exit 0, `Ran 4 tests ... OK` |
| V6 | Generated skill sections in sync | `python3 scripts/gen_skill_shortcut_sections.py --check` | none stated | PASS | Yes | exit 0, `visible_shortcuts=401 services=20` |
| V7 | No whitespace damage | `git diff --check upstream/main...HEAD` | remote named `upstream` | PASS | Yes | ran as `git diff --check <merge-base>...HEAD`, exit 0 (inferred equivalent — my remote is `origin`) |

No mismatches. Every claim reproduces with the stated conclusion.

## Behavioral and regression checks

- [PASS] `+report-latest` multi-page outbox — 2-page fixture now paginates and reads back the newest entry, where the previous revision hard-failed with `incomplete_latest_candidates` on any `hasMore` page. Genuinely regression-locking: the new `endTime`-snapshot assertion and the `cursor` 0→20 history assertion both fail against the merge-base implementation.
- [PASS] `+report-latest` bounded failure modes — stalled cursor, 10-page exhaustion, >200 items in one page, and conflicting-`createTime` duplicates each return a distinct structured error; identical-`createTime` overlap deduplicates and succeeds.
- [PASS] Sheet probe/write `sheetId` reuse — `waitSheetWritable` now returns the probed `sheetId`; the test asserts `calls = 4`, proving no extra `get_all_sheets`.
- [PASS] Bounded readback retry — `verifyRangesNotEmptyWithRetry` shares one backoff across probes and returns immediately on non-visibility errors (`TestVerifyRangeNotEmptyWithRetryStopsOnPermanentError` asserts `calls == 1` on `permission denied`).
- [FAIL] `addCreatedSheetID` result shape for `result`-wrapped responses — see finding 4.
- [FAIL] `report_received_today.py --days 1` at midnight — see finding 3.
- [PASS] `check-command-surface.sh --strict` — exit 0, `command surface check: ok`
- [PASS] `check-generated-drift.sh` — exit 0, `generated drift check: ok`, `schema assembly determinism check: ok` (registry_hash identical across both assembly passes)

### Finding 1 (blocking) — GFM table pipe escaping regression

`skills/multi/dingtalk-aitable/references/aitable.md` had 0 malformed table rows at merge-base and has 7 at HEAD, all on PR-added lines. The header declares two columns:

```
| 原子命令 | 用途 |
|---|---|
```

but these rows carry unescaped `|` inside the code span:

```
aitable.md:72 | `aitable view get aggregate|card|field-widths|...|visible-fields` | 读取单个视图配置面 |
aitable.md:73 | `aitable view update aggregate|card|...|visible-fields` | 更新一个已完整读取的视图配置面 |
aitable.md:75 | `aitable form field list|update|hide` | 表单字段顺序、展示和隐藏 |
aitable.md:76 | `aitable form questions create|delete` | 表单题目原子写入 |
aitable.md:77 | `aitable form share get|update` | 表单分享配置 |
aitable.md:79 | `aitable dashboard share get|update` | 仪表盘分享配置 |
aitable.md:81 | `aitable chart share get|update` | 图表分享配置 |
```

A code span does not protect `|` in GFM tables. Line 72 renders with 用途 = `card`, and `field-widths, fill-color-rule, filter, frozen-cols, group, lock, row-height, sort, timebar, visible-fields` vanish from the rendered page — the reader sees neither the subcommands nor the description. This is silent loss: nothing in CI fails.

The same table already uses the safe form three lines above (`| \`aitable view list\` / \`get\` | ... |`, lines 70–71, 74, 78, 80). Please either escape as `\|` or switch these rows to the ` / ` separator used by their neighbours.

### Finding 2 (blocking) — raw epoch timestamps contradict the contract this PR injects

`scripts/gen_skill_shortcut_sections.py` now injects the runtime contract into `references/report.md`, and that block states:

> `skills/multi/dingtalk-misc/references/report.md:15`
> - 时间戳面向用户展示时转换为带时区的可读时间；默认使用当前会话时区，必要时同时保留原值。

In the same PR, `report_received_today.py` stops formatting `createTime` and prints the raw value:

```python
# skills/multi/dingtalk-misc/scripts/report_received_today.py:279
f"{item.get('createTime', '')} | {item['reportId']}"
```

`createTime` from `+inbox-list --format json` is epoch milliseconds, so each line now shows a bare integer where the merge-base version printed `%Y-%m-%d %H:%M`. The PR adds the rule and breaks it in the same commit range. Either restore formatting (the script already builds a `ZoneInfo("Asia/Shanghai")` instant, so this is a one-line change) or keep the raw value alongside a readable one, as the contract's "必要时同时保留原值" allows.

### Finding 3 (blocking) — `--days 1` hard-fails during the first second after midnight

`report_received_today.py:243-248`:

```python
now = datetime.now(SHANGHAI)
start = (now - timedelta(days=args.days - 1)).replace(hour=0, minute=0, second=0, microsecond=0)
end = now.replace(microsecond=0)
```

With `--days 1`, `start` is today at 00:00:00 and `end` is `now` truncated to whole seconds. Between 00:00:00.000 and 00:00:00.999 the two are equal:

```
now=2026-08-26T00:00:00.500  start=2026-08-26T00:00:00+08:00  end=2026-08-26T00:00:00+08:00  end>start=False
now=2026-08-26T00:00:00.999  start=2026-08-26T00:00:00+08:00  end=2026-08-26T00:00:00+08:00  end>start=False
now=2026-08-26T00:00:01.000  start=2026-08-26T00:00:00+08:00  end=2026-08-26T00:00:01+08:00  end>start=True
```

`+inbox-list` rejects that window (`internal/shortcut/report/report.go:55`, `结束时间必须晚于开始时间`, reason `invalid_time_range`), so the script exits 2. The merge-base version pinned `end` to 23:59:59 and could never collide.

The window is one second wide, but it is the second in which a midnight-scheduled daily digest fires — which is the primary use case for a "reports received today" script. Clamping `end` to at least `start + 1s` fixes it.

### Finding 4 (blocking) — `addCreatedSheetID` mis-places `nodeId`/`sheetId` for `result`-wrapped responses

`parseCreatedNodeID` reads the create response through `unwrapSheetResult`, which unwraps a `result` envelope (`sheet_create_with_data.go:1519-1528`). The repository treats that shape as supported and tests it — `sheet_create_with_data_test.go:1314` asserts `{"result":{"nodeId":"N2"}}` → `N2`.

The new `addCreatedSheetID` does **not** unwrap; it unmarshals the raw text and injects both keys at the top level (`sheet_create_with_data.go:1406-1417`). I ran both shapes through the real functions:

```
IN   {"nodeId":"N2","docUrl":"https://x/y"}
  -> {"docUrl":"https://x/y","nodeId":"N2","sheetId":"SHEET_1"}          # correct

IN   {"result":{"nodeId":"N2","docUrl":"https://x/y"}}
  -> {"nodeId":"N2","result":{"nodeId":"N2","docUrl":"https://x/y"},"sheetId":"SHEET_1"}
```

In the wrapped case `nodeId` is duplicated across two levels and `docUrl` never reaches the top level — yet the `DataSchema` this PR adds (`sheet_create_with_data.go:128`) describes a flat object with `nodeId`, `sheetId`, `docUrl`. Because it declares `additionalProperties: true` and requires only `nodeId`/`sheetId`, schema validation still passes, so the result is silently mis-shaped rather than rejected.

The new test only covers `"{"` and `"null"` (`TestCrossPlatformCoverageAddCreatedSheetIDRejectsInvalidResponses`), so no test pins this. Please route the projection through `unwrapSheetResult` (or inject into the same object `parseCreatedNodeID` read from) and add a `result`-wrapped case.

### Finding 5 (non-blocking) — the headline Python window fix is not regression-locked

`report_received_today.py:247` carries a comment explaining the frozen window, but a mutation that restores the exact pre-PR `end = now.replace(hour=23, minute=59, second=59)` still passes the new suite. The same is true for the cursor strict-advance check, `max_pages` exhaustion, the `MAX_REPORTS` cap, the total timeout, and `SHANGHAI` → naive local.

Worth contrasting with the Go side, which does lock its equivalent change: `TestCrossPlatformCoverageReportLatestExecutorRemainingBranches` asserts `endTime` falls inside `[before, after]` of the invocation, which fails against the merge-base end-of-day window. A comparable assertion on the Python `--dry-run` path would close this.

### Finding 6 (non-blocking) — contracts dropped by the compression

The Summary says the compression preserves "success-critical contracts". Most did survive (Asia/Shanghai, `--to-user-ids` mandatory + invisible-on-success, the 20-day window, strict-forward cursor, `--styles` key order and non-atomicity, `--ranges` max 100, the size-type enum). These did not, and each is enforced in Go:

1. **`--size` ceiling of 20** — validated at `internal/shortcut/report/report.go:40,99,105`. HEAD says "默认每页 `--size 20`" but no longer that 20 is the maximum; the previous text explicitly forbade 50/100.
2. **`contents` 10MB limit** — `internal/helpers/report.go:24` (`reportContentsMaxBytes`), error raised at `:1736-1737`. Gone from `references/report.md` even though HEAD still routes to `entry submit`. Still present in `skills/mono/references/products/report.md:140,371,383`, so the fact now lives in one tree only.
3. **Sheet chart field reference** — the `title`/`legend`/`catAx`/`valAx` tables were deleted (`axisMin`, `axisMax`, `splitLine`, `minorSplitLine`, `axisLabel`, `axisLine`, `pos` enums, `titleConfig`) with no equivalent anywhere on HEAD.
4. **`--styles` unknown-key rejection** — the `styleItemFields` whitelist (`sheet_create_with_data.go:751`) errors at `:1011`.

Either restore these or record them as intentional scope cuts under T8.

### Finding 7 (non-blocking) — `--days N` above 200 reports now fails instead of degrading

`HARD_MAX_PAGES = 10` caps `--max-pages` with no override, and exceeding `MAX_REPORTS` raises. A `--days 30` window with more than 200 received reports now exits 2 where the merge-base version returned a truncated first page. Fail-closed is defensible, but it is a user-visible capability change and belongs in the fragment.

### Observations (not defects, no action requested)

- `skills/mono/references/products/report.md` is untouched while its multi counterpart went 537→103 lines, widening divergence from 188 to 519 diff lines. The two document different surfaces (mono teaches legacy `report outbox list`, multi teaches `+` shortcuts), so they are not strict copies, and `skills/mono/scripts/report_received_today.py` already differed at merge-base.
- `test/scripts/release_script_test.go` disables `maintenance.auto`/`gc.auto` in test repos. The rationale is sound and well-commented, and it is a genuine flake fix rather than masking — but it sits outside the declared Scope.
- The PR **removes** 10 references to a competitor product from the Sheet docs. Four remain in Go help text (`sheet_create_with_data.go:63,76,159,744`, with line 159 reaching users through `--help`); those lines are untouched here and are pre-existing.
- `rt.Command().SetContext(...)` in `latest.go:68` is safe — `internal/app/root.go:842` already sets the command context per execution — but `+report-latest` becomes the only shortcut with a hardcoded 2-minute self-imposed timeout. Worth a note if that is meant to generalise.
- `skills/multi/dingtalk-misc/SKILL.md` drops the CRITICAL "read `dingtalk-shared` before any `dws` operation" precondition for Report/Sheet. The generator change does inject the equivalent contract into both references, so the substitution is wired up correctly — flagging only because it is a safety-baseline routing change that the Summary does not mention.

## Release fragment validation

- [FAIL] F1 Fragment present and legal — no file added under `.changes/`. The PR carries several user-visible changes: `+report-latest` goes from hard-failing on any multi-page outbox to bounded pagination; `sheet create-with-data` now publishes `sheetId` in its result and declares a `DataSchema`; `report_received_today.py` changes its output line format, its timestamp rendering, and its exit code (0 → 2 on error). The closest precedent by file composition is `.changes/report-entry-submit-require-recipient.md` (`category: Changed`), which was added for a comparable Report CLI behavior tightening. `.changes/README.md` puts the call with the reviewer — "评审者根据改动是否可见来判断该例外是否成立" — and on this diff the changes are visible.
- [N/A] F2 Category valid — no fragment to grade. When you add one, `Changed` fits the Report/Sheet behavior shift; `Fixed` fits if you frame it as the pagination bug.
- [N/A] F3 Body well-formed — no fragment to grade
- [PASS] F4 `CHANGELOG.md` untouched — confirmed, correct for an ordinary PR
- [N/A] F5 English-only, no rival names — no fragment to grade. Please write it in **English**: 18 of the 20 fragments under `.changes/` and `.changes/released/` contain zero CJK characters, so English is the established convention. Do not copy the Chinese style of `report-entry-submit-require-recipient.md` (82 CJK) — that one is the exception, not the pattern.
- [N/A] F6 Fragment matches the diff — no fragment to grade
- [PASS] `check-release-fragments.sh <merge-base> HEAD` — exit 0 with **no output**. Note this gate does not require a fragment to exist, so the green `Policy` check is not evidence for F1.
- Baseline: 0 rival-product names and 0 CJK characters introduced by this PR into any fragment (there is no fragment); the Sheet docs remove 10 pre-existing rival references.

## Conventional code review

- [PASS] Diff scope and surrounding call paths — all 34 files read; `unwrapSheetResult`, `parseCreatedNodeID`, `reportProjectEntries`, `reportValidateContinuation`, `reportOptionalInteger`, `reportLatestEntryID` traced
- [PASS] Error handling and failure paths — Report returns distinct structured reasons (`stalled_cursor`, `conflicting_duplicate_item`, `latest_item_limit_reached`, `latest_page_limit_reached`); Sheet separates convergent from permanent errors via `errSheetReadbackNotVisible`, which is the right seam
- [PASS] Concurrency, cancellation, cleanup, lifecycle — `waitSheetRetryDelay` now honours `ctx.Done()` and `runCreateSheetWithData` inherits `cmd.Context()` instead of `context.Background()`, both improvements; the removed "此处不响应 ctx 取消" comment is superseded intentionally
- [PASS] Security, permissions, temporary files, command execution — no new command execution, no credential handling, no temp files
- [FAIL] Backward compatibility and configuration defaults — `report_received_today.py` changes its output line format, timestamp rendering, and exit code without a fragment; `create-with-data` re-marshals the create response (keys are now alphabetically sorted rather than server order — harmless to JSON parsers, noted for completeness)
- [PASS] Platform-specific files and build tags — none changed; the `TestCrossPlatformCoverage` prefix is used correctly on the new tests so the platform coverage gate stays active
- [FAIL] Test coverage quality — strong on the Report Go path (see Behavioral checks). Two gaps: no `result`-wrapped case for `addCreatedSheetID` (finding 4) and the Python suite does not lock its headline fix (finding 5)
- [PASS] Formatting, static analysis, generated files, dependency scope — `git diff --check` clean; drift and command-surface both green; no dependency changes
- [FAIL] Documentation consistency with the code — findings 1, 2 and 6

## CI and baseline comparison

- [PASS] Required CI checks — all 42 rollup entries are `SUCCESS` or `SKIPPED`, none failing. Verified the three runs (`32917470468`, `32917469000`, `32917468875`) all report `headSha=76fd76d08bc11f9650d703ed3dacc61e38dcec7a`, i.e. they belong to the current head rather than a stale SHA.
- [N/A] Claimed known failures reproduced on PR — the only claimed failure is the host killing the local Schema Catalog generator with signal 9, an environment limit rather than a code failure. The narrower `TestReviewedRoutedInterfacesReachFinalSchema` passes here (V4), and `check-generated-drift.sh` completes successfully in this environment including the determinism double-assembly.
- [N/A] Base comparison — no mismatch required classification; every reproduced command matched the author's conclusion

## Findings and blockers

- Blocking code findings:
  - `skills/multi/dingtalk-aitable/references/aitable.md:72,73,75,76,77,79,81` — unescaped `|` truncates 7 table rows (finding 1)
  - `skills/multi/dingtalk-misc/scripts/report_received_today.py:279` — raw epoch millis contradicts `references/report.md:15` (finding 2)
  - `skills/multi/dingtalk-misc/scripts/report_received_today.py:243-248` — `--days 1` exits 2 in the first second after midnight (finding 3)
  - `internal/helpers/sheet_create_with_data.go:1406-1417` — `nodeId`/`sheetId` injected at the wrong level for `result`-wrapped responses (finding 4)
- Description blockers (template-required only): T2, T3, T4, T5, T6, T7, plus the false `skills/mono` Scope claim
- Fragment blockers: F1
- Validation mismatches: none — V1 through V7 all reproduce
- Not-verified items: none material. Live MCP behavior against the real Report/Sheet backends was not exercised (no credentials), which is a property of every tool binding in this repository and not specific to this PR.

## Re-review condition

1. Correct the `## Scope` bullet so it reflects the 11 changed `skills/mono/` files.
2. Add `.changes/<name>.md` (English, `category: Changed` or `Fixed`) covering: `+report-latest` bounded pagination, the new `sheetId` in the `create-with-data` result, and the `report_received_today.py` output/exit-code change. Or argue the exception explicitly if you disagree.
3. Fill in T2 (risk tier), and give T5/T6/T7 results — the commands you already ran are sufficient, they just need their outcomes recorded.
4. Fix findings 1–4.
5. For findings 5–7, either address them or record them as intentional scope cuts under `## Notes`.

I will re-run V1–V7, the two conditional gates, the `addCreatedSheetID` shape probe, and the midnight window probe against the new head.

pengzhihan47-star:
@haofeng0705 Thanks for the detailed review. I addressed the blocking findings and the requested description/release items in `41bebea6`.

### Blocking findings

1. **GFM table truncation** — replaced the seven unescaped inline `|` command separators with separate code spans joined by ` / `; every affected row now has exactly two rendered columns.
2. **Raw Report timestamps** — added `format_create_time` so epoch milliseconds render as a readable Shanghai-time value with an explicit `+0800` offset.
3. **Midnight `--days 1` failure** — extracted `query_window` and clamps `end` to at least `start + 1s`. `test_query_window_clamps_first_second_after_midnight` reproduces the 00:00:00.500 case.
4. **Wrapped Sheet result shape** — `addCreatedSheetID` now projects from the same unwrapped business object used by `parseCreatedNodeID`. `TestCrossPlatformCoverageAddCreatedSheetIDUnwrapsBusinessResult` verifies a flat `nodeId` / `sheetId` / `docUrl` result, no envelope leakage, and exact preservation of a >2^53 integer.

### Non-blocking findings

- Added regression coverage for the frozen current-time window and readable timezone formatting (finding 5).
- Restored the four compressed success-critical contracts: Report `--size <= 20`, Report contents <= 10MB, Sheet chart title/legend/axis fields, and `create-with-data --styles` unknown-key rejection. The two changed Sheet sub-references were mirrored to `skills/mono` as required (finding 6).
- Kept the 10-page / 200-item behavior intentionally fail-closed so partial results cannot be presented as complete; this scope cut is now explicit under PR Notes and in the release fragment (finding 7).

### PR metadata and verification

- Corrected Scope: the 11 `skills/mono` files are declared as required Sheet mirrors; there are still no `dingtalk-doc` product changes.
- Added the Standard risk tier, release fragment `.changes/1136-report-sheet-workflow-correctness.md`, release-seal N/A reason, test outcomes, behavior evidence, and conditional gate results.
- Local affected-package verification passes, including the full helpers suite, Report tests, build, final-Schema contract test, 7 Python regressions, generated Skill-section check, and diff check.
- Current head `41bebea6` is not behind main and all 38 GitHub checks are green, including Policy and Interface Integrity.

Please re-review when convenient.

maoqxxmm:
## Sheet 业务影响评估（基于 `41bebea`）

结论：**建议先修后合**。本 PR 没有直接修改 Sheet 后端、HSF 或 MCP 接口；`create-with-data` 的运行时调整整体是正向的，但新的 Agent 路由文档存在 2 个会影响核心能力可发现性/正确路由的问题，另有 1 个数据验证契约被过度压缩的问题。

### 正向影响

- `create-with-data --format json` 现在直接输出扁平业务对象，并在顶层补充探测得到的 `sheetId`；该字段来自 CLI 的 `get_all_sheets`，不是后端新增字段。
- 复用探测到的默认 `sheetId`，典型单 Sheet `--values` 链路调用数约从 5 次降到 4 次，多 Sheet 约从 8 次降到 7 次。
- 命令上下文已正确向下传递；写后回读仅在“读到空数据”时重试，退避为 0/250/500/1000 ms，最多读 4 次，不会重放写请求，永久错误会立即停止。
- 输出投影保留业务字段和大整数，并兼容解包 `result`，最新修复已覆盖此前 review 中的 response flatten、图表轴/图例字段和 `--styles` 未知字段校验问题。

### 合入前建议修复

1. **[P1] 本地 Excel 导入路由互相矛盾**

   根路由把 `xlsx/xls/xlsm/csv` 统一描述为“下载到本地处理；不要对二进制文件调用 Sheet 子命令”，但导入子文档又明确规定“把本地 xlsx/xls 转为在线表格”应走 `dws sheet import create`。Agent 可能因此把核心导入场景错误路由成本地处理。

   建议按意图区分：本地 `xlsx/xls` 转在线表格 → `sheet import create`；Drive 二进制转在线表格 → 先 `drive download` 再 `sheet import create`；仅本地分析才走本地处理；`xlsm/csv` 单独说明不属于当前 import 支持范围。

   证据：[根路由](https://github.com/DingTalk-Real-AI/dingtalk-workspace-cli/blob/41bebea638b0ca8e877af0a0ae0784a133cbc114/skills/multi/dingtalk-misc/references/sheet.md#L21-L30) / [import 契约](https://github.com/DingTalk-Real-AI/dingtalk-workspace-cli/blob/41bebea638b0ca8e877af0a0ae0784a133cbc114/skills/multi/dingtalk-misc/references/sheet/sheet-import.md#L1-L10)

2. **[P1] Sheet 模板能力失去路由入口**

   CLI 仍正式提供 `sheet template list/search/apply`，但新的复杂阶段索引没有 template 行，新的 Sheet reference 也没有“模板/template”入口。由于新说明同时弱化了父级 Help/Catalog 探索，Agent 很难再发现这组能力。

   建议恢复一条 `template list/search/apply` 阶段路由，或补一个最小 `sheet-template.md`。

   证据：[CLI 模板命令](https://github.com/DingTalk-Real-AI/dingtalk-workspace-cli/blob/41bebea638b0ca8e877af0a0ae0784a133cbc114/internal/helpers/sheet_template.go#L10-L190) / [新阶段索引](https://github.com/DingTalk-Real-AI/dingtalk-workspace-cli/blob/41bebea638b0ca8e877af0a0ae0784a133cbc114/skills/multi/dingtalk-misc/references/sheet.md#L103-L128)

3. **[P2] `dataValidation` 三态契约被过度压缩**

   新文档只保留“内联选项 vs `sourceRange`”，遗漏了 `checkbox`，以及“不传 `dataValidation` 表示保留、`{"type":"none"}` 表示清除”的关键差异，可能导致 Agent 无法创建复选框或误判清除方式。

   建议恢复一条紧凑说明，覆盖 `dropdown/checkbox/none` 与 preserve/clear 语义。

   证据：[新写入文档](https://github.com/DingTalk-Real-AI/dingtalk-workspace-cli/blob/41bebea638b0ca8e877af0a0ae0784a133cbc114/skills/multi/dingtalk-misc/references/sheet/sheet-write-data.md#L25-L38) / [实际命令契约](https://github.com/DingTalk-Real-AI/dingtalk-workspace-cli/blob/41bebea638b0ca8e877af0a0ae0784a133cbc114/internal/helpers/sheet_range_ops.go#L160-L177)

以上 import/template 问题在 mono 与 multi 两套 Sheet skill 中都存在。

### 兼容性与运行风险

- `create-with-data --format json` 从 transport wrapper 改为扁平业务对象；依赖 `.result.nodeId`、外层 `requestId`、精确 JSON 字节或 key 顺序的脚本会受影响。过渡期消费者建议同时接受顶层对象和旧 `result` wrapper。
- 每个非空 Sheet 最坏会有 4 次读（3 次额外重试），共享总退避上限 1.75 秒；建议观察多 Sheet 场景的读 QPS 与 P95。
- 复合流程仍非原子：create/write 成功后 styles 失败时，应复用返回的 `nodeId` 继续处理，不能整体重试 create，以免生成重复工作簿。

### 已验证与待补验证

已通过：

- `go test ./internal/helpers -count=1`
- 本 PR 新增的 Sheet create/readback/context/retry 聚焦测试
- `TestReviewedRoutedInterfacesReachFinalSchema`
- `python3 scripts/gen_skill_shortcut_sections.py --check`
- `git diff --check`

合入前建议再补：

- 针对“本地 Excel 导入、模板 list/search/apply、checkbox/清除 dataValidation”的静态路由测试。
- 基于 base/head 的真实模型 Agent A/B，多轮覆盖上述 Sheet 意图。
- 专用测试工作簿上的真实 CLI → MCP `create-with-data` 单/多 Sheet 写入与独立回读，确认 `sheetId`、无重复创建及实际调用量/延迟。

完成 P1 修复并补齐对应路由验证后，再合入风险更可控。

haofeng0705 (changes requested):
## Gate decision

REQUEST CHANGES — findings 1–3 and 5–7, the Scope correction, the description gaps and the release fragment are all closed and independently verified. Two new blockers arrived with the fix commit itself: the finding-6 chart restoration reintroduces the finding-1 pipe defect in two files, and the finding-4 fix over-narrows so that four response shapes the previous head handled *correctly* now hard-fail after the workbook was already created. A third item — three compression regressions of the finding-6 class raised by @maoqxxmm — reproduces here and is not recorded under `## Notes`.

The work on the four original blockers is good: every fix reproduces, and both new Go and Python regression tests fail against the pre-fix head when ablated, so they genuinely pin their bugs rather than passing incidentally.

## Review provenance

- [PASS] Analysis performed at `41bebea638b0ca8e877af0a0ae0784a133cbc114` in a clean dedicated detached checkout
- [PASS] **Head drifted during the review**: review SHA = `d04b7dc2c3561981e77fa639e6e7cd807dc30144`, evidence SHA = `41bebea6`. The delta is a `main` merge (bringing in #1148's draft-CI gates) touching six files — `.github/workflows/{ai-behavior-check,ci,draft-ci}.yml`, `docs/ci-pr-gates.md`, `test/scripts/{draft_pr_ci_workflow,reviewer_router_workflow}_test.go` — none of which this PR owns. I hashed every file the PR touches at both heads: **0 differ**, so all conclusions below stand unchanged. Re-confirmed at `d04b7dc2` directly: the whole-corpus malformed-row count is still 14, both `sheet-chart.md:45` rows still carry the unescaped pipes, the `addCreatedSheetID` rejection branch is unchanged, and `sheet template` still has 0 routes.
- [PASS] Base revision / merge-base `76618a6b8c73dac98935da047d65dad5d892e72f` — unchanged across both heads and identical to the previous review, so no base drift
- [PASS] Previous review head `76fd76d08bc11f9650d703ed3dacc61e38dcec7a` is an ancestor of `41bebea6` (no force-push), so `git log`/`git diff` across the two are meaningful; the fix delta is one commit, 11 files, +127/−21

## PR description vs repository template

Template graded against `.github/pull_request_template.md`.

- [PASS] T1 Summary — four bullets state what changed and why; the fourth now names the GFM repair
- [PASS] T2 Risk tier — `Standard` selected, and it matches the diff (implementation change, stable package graph, no workflow/packaging/auth surface)
- [PASS] T3 Release fragment — `.changes/1136-report-sheet-workflow-correctness.md`
- [PASS] T4 Release-seal validation — `N/A` with a reason that holds (ordinary PR, `CHANGELOG.md` untouched)
- [PASS] T5 Targeted test/check commands and results — all seven now carry outcomes, and all seven reproduce (table below)
- [PASS] T6 Behavior evidence — three test names given; I ablated all three (see Behavioral checks)
- [PASS] T7 Conditional checks — the two substituted claims are correct, and I confirmed them directly at this head rather than relying on the preceding-head result: `check-command-surface.sh --strict` exit 0 (`command surface check: ok`), `check-generated-drift.sh` exit 0 (`generated drift check: ok`, `schema assembly determinism check: ok`, `registry_hash=sha256:5f4103da…` identical across both assembly passes). `gen_skill_shortcut_sections.py --check` exit 0 confirms the hand edits to `references/report.md` sit outside the generated markers.
- [PASS] T8 Notes — finding 7 is recorded as an intentional fail-closed scope cut, and the four restored contracts are listed

**Scope claim now correct.** `git diff --name-only <merge-base>...HEAD -- skills/mono/` returns exactly **11** files, matching the corrected bullet. The `dingtalk-doc` half remains accurate.

One observation on T7, not a defect in the claim: "every AITable View/Form/Dashboard/Chart table row now has exactly two GFM columns" is literally true and I verified it. The rendering check was scoped to AITable only, which is why the two malformed rows the *same commit* added to the Sheet chart references (finding A) were not caught.

Observations outside the template — explicitly **not** counted as defects: local OS/Go version, a requirement-to-test mapping table, a dedicated limitations section, reproducing the full CI suite locally, and upstream MCP contract verifiability.

## Author validation reproduction

| ID | Acceptance criterion | Author claim / exact step | Documented environment | Independent result | Match | Evidence |
|---|---|---|---|---|---|---|
| V1 | Build succeeds | `make build` | none stated | PASS | Yes | exit 0 |
| V2 | Sheet helpers green | `go test ./internal/helpers -count=1` | none stated | PASS | Yes | `ok … 105.378s` (author: 103.064s) |
| V3 | Report shortcuts green | `go test ./internal/shortcut/report -count=1` | none stated | PASS | Yes | `ok … 0.779s` |
| V4 | Routed interfaces reach final Schema | `go test ./internal/app -run TestReviewedRoutedInterfacesReachFinalSchema -count=1` | none stated | PASS | Yes | `ok … 1.007s` |
| V5 | Python suite green, 7 tests | `python3 scripts/dev/report_received_today_test.py` | none stated | PASS | Yes | `Ran 7 tests … OK` (was 4 at the previous head) |
| V6 | Generated skill sections in sync | `python3 scripts/gen_skill_shortcut_sections.py --check` | none stated | PASS | Yes | exit 0, `visible_shortcuts=401 services=20` |
| V7 | No whitespace damage | `git diff --check upstream/main...HEAD` | remote named `upstream` | PASS | Yes | ran as `git diff --check <merge-base>...HEAD`, exit 0 (inferred equivalent — my remote is `origin`) |

No mismatches. Every claim reproduces with the stated conclusion.

## Behavioral and regression checks

### Previous blockers — verified closed

- [PASS] **Finding 1 — GFM pipes in `aitable.md`.** Malformed table rows in that file: base **0** → previous head **7** → this head **0**. All seven rows now use the ` / ` separator their neighbours already used.
- [PASS] **Finding 2 — raw epoch timestamps.** `format_create_time` now renders readable Shanghai time with an explicit offset. Measured directly against the packaged module:

  ```
  1756180000000    -> '2025-08-26 11:46:40 +0800'
  0                -> '1970-01-01 08:00:00 +0800'
  '1756180000000'  -> '1756180000000'      # non-numeric shape kept verbatim
  None / ''        -> ''
  True             -> 'True'               # bool correctly excluded from the numeric branch
  ```

  The string passthrough is consistent with the contract's "必要时同时保留原值".
- [PASS] **Finding 3 — midnight `--days 1`.** `query_window` clamps `end` to at least `start + 1s`. Ablated on a copy of the packaged script:
  - remove the clamp (restore the previous head's bug) → `FAILED (failures=1)`
  - restore the base `end = now.replace(hour=23, minute=59, second=59)` → `FAILED (failures=2)`
  - unmutated control → `Ran 7 tests … OK`

  So the new tests pin both the midnight collision *and* the frozen-window property.
- [PASS] **Finding 4 — `result`-wrapped create response.** Fixed for the reported shape, and the new test has real power. I copied `TestCrossPlatformCoverageAddCreatedSheetIDUnwrapsBusinessResult` verbatim onto the previous head: it fails there with exactly the value I reported — `{"nodeId":"probed-node","requestId":"outer","result":{…},"sheetId":"probed-sheet"}`. See finding B for the over-narrowing this fix introduced.
- [PASS] **Finding 5 — headline Python fix now regression-locked.** The exact mutation I named in that finding (`end = now.replace(hour=23, minute=59, second=59)`) is now caught, as is a `SHANGHAI` → naive-local mutation via the `+0800` assertion. Residual, matching your scoped reply and not requested: the strict-cursor advance, `max_pages` exhaustion, the `MAX_REPORTS` cap and the total timeout are still not mutation-locked on the Python side.
- [PASS] **Finding 6 — four contracts restored.** Report `--size` ceiling of 20 (`references/report.md:63`, including the explicit 50/100 prohibition), the `contents` 10MB limit (`:84`), the chart `title`/`legend`/`catAx`/`valAx` field set, and the `--styles` top-level key whitelist with pre-create rejection. Both changed Sheet sub-references are byte-identical between `skills/mono/references/products/sheet/` and `skills/multi/dingtalk-misc/references/sheet/`. Content restored — but see finding A for how it renders.
- [PASS] **Finding 7 — fail-closed above 10 pages / 200 reports** recorded under `## Notes` and reflected in the fragment ("fails closed instead of returning incomplete data").

### Finding A (blocking) — the chart restoration reintroduces the finding-1 pipe defect

`skills/multi/dingtalk-misc/references/sheet/sheet-chart.md:45` and `skills/mono/references/products/sheet/sheet-chart.md:45` (identical files):

```
| 对象 | 已知字段与枚举 |
|---|---|
...
| `catAx` / `valAx` | `show:boolean`、`pos:l/t/b/r`、`titleConfig:{show,title}`、`axisMin:number|null`、`axisMax:number|null`、`splitLine:boolean`、`minorSplitLine:boolean`、`axisLabel:boolean`、`axisLine:boolean` |
```

The header declares two columns; `axisMin:number|null` and `axisMax:number|null` carry unescaped pipes, so the row splits into four cells and GFM discards the last two:

```
cell 1 [RENDERED] `catAx` / `valAx`
cell 2 [RENDERED] `show:boolean`、`pos:l/t/b/r`、`titleConfig:{show,title}`、`axisMin:number
cell 3 [DROPPED]  null`、`axisMax:number
cell 4 [DROPPED]  null`、`splitLine:boolean`、`minorSplitLine:boolean`、`axisLabel:boolean`、`axisLine:boolean`
```

So on the rendered page `axisMax`, `splitLine`, `minorSplitLine`, `axisLabel` and `axisLine` disappear and `axisMin` reads as plain `number` — those are precisely the fields finding 6 item 3 asked you to restore. Whole-corpus count over all 350 skill Markdown files, header-column vs row-cell mismatch: base **12** → previous head **19** (+7, finding 1) → this head **14** (−7 fixed, **+2 new**). The row also ships inside the built binary (one occurrence in `dws`), so it reaches installed skill trees, and nothing in CI fails.

The `legend` row two lines above already uses the safe form (`` `t` / `b` / `l` / `r` / `none` ``). Either escape as `axisMin:number\|null` or split into separate code spans joined by ` / `, in both files.

### Finding B (blocking) — `addCreatedSheetID` now hard-fails on shapes the previous head handled correctly

The fix unwraps whenever the `result` **key** exists and errors when its value is not an object. `unwrapSheetResult` — used by `parseCreatedNodeID` and three other call sites in the same file — unwraps only when `result` **is** an object and otherwise returns the outer object. Differential probe calling the real functions, previous head vs this head:

| create response | `parseCreatedNodeID` | `addCreatedSheetID` @ `76fd76d0` | `addCreatedSheetID` @ `41bebea6` |
|---|---|---|---|
| `{"result":{"nodeId":"N2","docUrl":"…"}}` | `N2` | envelope leaked (the reported bug) | flat, correct |
| `{"nodeId":"N2","docUrl":"…","result":null}` | `N2` | flat, correct | **ERR** `创建响应 result 不是 JSON 对象` |
| `{"nodeId":"N2","docUrl":"…","result":[]}` | `N2` | flat, correct | **ERR** |
| `{"nodeId":"N2","docUrl":"…","result":"ok"}` | `N2` | flat, correct | **ERR** |
| `{"nodeId":"N2","docUrl":"…","result":0}` | `N2` | flat, correct | **ERR** |

Three things make this worth fixing rather than accepting as fail-closed:

1. It contradicts the invariant the fix's own comment states — "MCP 返回 result 包装时，投影必须与 parseCreatedNodeID 读取同一层". On all four shapes `parseCreatedNodeID` still succeeds by reading the outer object while the projection refuses to read any level.
2. The projection runs at `internal/helpers/sheet_create_with_data.go:335`, after the workbook was created, the initial data written *and* read back, and `--styles` applied. The command therefore returns `表格已创建且初始数据已写入 (nodeId=…)，但构造结果失败: …` for an operation that fully succeeded, and the probed `sheetId` never reaches the caller. Failing closed protects nothing at that point — the writes already landed; it only destroys the receipt.
3. The four shapes were handled *correctly* before this commit. The bug was confined to the object-wrapped case; the fix corrected that one and broke four that were already right.

`TestCrossPlatformCoverageAddCreatedSheetIDRejectsInvalidResponses` now pins the new behavior by listing `{"result":null}` and `{"result":[]}` as required rejections, so this will not self-correct. Suggested fix: apply `unwrapSheetResult`'s rule — unwrap only when `result` decodes to an object, otherwise project into the outer object — and move those two shapes from the reject list to the accept list.

### Finding C (blocking) — three finding-6-class compression regressions, none recorded under `## Notes`

@maoqxxmm raised these; all three reproduce here, and all three are introduced by this PR. Finding 6 asked for "either restore these or record them as intentional scope cuts under T8", and `## Notes` currently lists only the four restored contracts plus the fail-closed pagination.

**C1 — the product-boundary table now misroutes local-Excel import.** Base had a discriminating row:

```
| 本地 xlsx/xls 导入为在线表格 | dws sheet import create --file <路径> … |
| `xlsx` / `xls` / `xlsm` / `csv`  | dws drive download … 禁止调用任何 sheet 子命令 |
```

HEAD collapses both into one row: `| 本地或 Drive 中的 xlsx/xls/xlsm/csv 文件 | 下载到本地处理；不要对二进制文件调用 Sheet 子命令 |`. But `sheet import create` *is* a Sheet subcommand, and `sheet/sheet-import.md` — which this PR does not change — still teaches it as the path for "把 xlsx 变成在线表格". The file itself says "先用本文件完成路由", so the boundary table is consulted before the stage index, and the local-xlsx→online-sheet intent is steered away from the correct command. The stage index still reaches `sheet-import.md` via the "导入或导出边界/失败恢复" row, so the reference is not orphaned — but the first routing step contradicts it. Splitting by intent, as @maoqxxmm suggests, resolves it.

**C2 — `sheet template` lost its only route.** `dws sheet template --help` on the head build lists `apply` / `list` / `search`, so the capability is live. Mentions across the whole skills corpus: base **2 files** (`skills/multi/dingtalk-misc/references/sheet.md` with a dedicated `| sheet-template | 表格模板管理…命令：template list/search/apply |` stage row plus embedded help for all three subcommands, and `skills/mono/references/products/sheet.md`) → HEAD **0 files out of 350**. Compounding it, this PR also adds "不要加载产品级 Catalog 代替选路" and "禁止通过父级 Help … 或完整 Catalog 探索命令", so the fallback discovery path is explicitly closed at the same time.

**C3 — `dataValidation` tri-state and `checkbox` dropped, asymmetrically.** Base documented all three semantics explicitly (omit → preserve; `{"type":"none"}` → clear; `dropdown`/`checkbox` → overwrite). HEAD keeps only "内联选项和 sourceRange 二选一". `checkbox` mentions in the multi Sheet references: base **11** → HEAD **0** (and 0 in the mono mirrors too). Two lines above in the same bullet list, the parallel `hyperlink` tri-state *survives* ("取消整格链接使用 `hyperlink:{"type":"none"}`；不传 hyperlink 表示保留原链接"), and `internal/helpers/sheet_range_ops.go:165-177` documents all three plus `checkbox` for `dataValidation` — which is why this reads as an oversight rather than a deliberate cut.

## Release fragment validation

- [PASS] F1 Fragment present and legal — `.changes/1136-report-sheet-workflow-correctness.md`, flat at the top level, regular file, name matches `^[a-z0-9][a-z0-9._-]*\.md$`
- [PASS] F2 Category valid — `Changed`
- [PASS] F3 Body well-formed — three Markdown list items, no `TODO`/`TBD`
- [PASS] F4 `CHANGELOG.md` untouched — the only `.changes/`-or-`CHANGELOG` path in the diff is the new fragment
- [PASS] F5 English-only, no rival product names — 0 CJK characters and 0 rival-name hits on the added fragment lines; 0 rival hits across every line this PR adds anywhere, and the PR removes 10 pre-existing rival references from the Sheet docs
- [PASS] F6 Fragment matches the diff — all three bullets check out against the code: bounded strictly-advancing pagination with duplicate reconciliation and exact readback; the probed `sheetId` in a flat declared result shape with bounded readback and no repeated `get_all_sheets`; and the helper's bounded pagination, Shanghai timestamps, fail-closed behavior and valid midnight window
- [PASS] `check-release-fragments.sh <merge-base> HEAD` — exit 0, no output. As before, this gate does not require a fragment to exist and does not inspect language, so it is not evidence for F1 or F5.
- Baseline: 0 CJK and 0 rival names introduced into any fragment.

One optional improvement, not a FAIL: at base, `create-with-data` ended in `PrintRaw(createText)` — the raw MCP response verbatim. For `result`-wrapped responses the output is now the flattened business object with the transport envelope (including `requestId`) removed. "with a flat … result shape" does signal this, but naming the envelope removal explicitly would help consumers that read `.result.nodeId` or the outer `requestId`.

## Conventional code review

- [PASS] Diff scope and surrounding call paths — the full one-commit delta read; `unwrapSheetResult`, `parseCreatedNodeID`, `addCreatedSheetID`, `query_window`, `format_create_time` and the `projectCreatedSheetResult` call site traced
- [PASS] Error handling and failure paths — the Report structured reasons and the `errSheetReadbackNotVisible` seam are unchanged and still correct
- [FAIL] Concurrency, cancellation, cleanup, or lifecycle — finding B: the terminal projection can fail an already-completed create/write/style sequence, leaving the caller without the `sheetId` for a workbook that exists
- [PASS] Security, permissions, temporary files, and command execution — no new command execution, credentials or temp files
- [PASS] Backward compatibility and configuration defaults — the fragment now covers the user-visible changes; the envelope-stripping note above is the only residual
- [PASS] Platform-specific files and build tags — none changed; the new Go test keeps the `TestCrossPlatformCoverage` prefix, so the platform coverage gate stays active
- [PASS] Test coverage quality — both previously-flagged gaps are closed, and both new tests survive ablation against the previous head. The one remaining concern is that the Go test pins the *wrong* rejection set (finding B).
- [PASS] Formatting, static analysis, generated files, and dependency scope — `git diff --check` clean; command-surface and generated-drift both green locally at this head; no dependency changes
- [FAIL] Documentation consistency with the code — findings A and C

## CI and baseline comparison

- [PASS] Required CI checks — CI run `32927145274` completed `success` on `41bebea6`, so your "all checks green" statement was accurate at the time of your comment. Note for anyone reading later: a fresh run (`32928509028`) is in flight on the same SHA at review time, with 24 checks `QUEUED`/`IN_PROGRESS`; three earlier runs on this SHA are `cancelled` by cancel-in-progress, which is not a gate failure.
- [PASS] Claimed known failures reproduced — the only claimed failure remains the local host killing the Schema Catalog generator with signal 9, an environment limit. In this environment `check-generated-drift.sh` completes, including the determinism double-assembly, so it is host-specific rather than a code failure.
- [PASS] Base comparison — merge-base unchanged at `76618a6b`, so no base drift; every differential probe was run on previous head and this head, and the previous head reproduced the values recorded in my earlier review, confirming the probes still discriminate.

## Findings and blockers

- Blocking code findings:
  - `skills/multi/dingtalk-misc/references/sheet/sheet-chart.md:45` and `skills/mono/references/products/sheet/sheet-chart.md:45` — unescaped `|` drops 5 of the 9 restored axis fields from the rendered table (finding A)
  - `internal/helpers/sheet_create_with_data.go:1415-1421` — non-object `result` hard-fails a fully-completed create, diverging from `unwrapSheetResult` and from the fix's own stated invariant (finding B)
  - `skills/multi/dingtalk-misc/references/sheet.md` (product-boundary table, stage index) and `skills/multi/dingtalk-misc/references/sheet/sheet-write-data.md` plus their mono mirrors — import misrouting, the lost `sheet template` route, and the dropped `dataValidation` tri-state / `checkbox` (finding C)
- Description blockers (template-required only): none — T1 through T8 all pass and the Scope claim is now accurate
- Fragment blockers: none — F1 through F6 all pass
- Validation mismatches: none — V1 through V7 all reproduce
- Not-verified items: live MCP behavior against the real Report/Sheet backends (no credentials), which is a property of every tool binding in this repository and not specific to this PR. I could not confirm whether the upstream `create_workspace_sheet` ever returns a non-object `result`; finding B rests on the internal divergence and the post-write failure mode rather than on an observed upstream shape.

## Re-review condition

1. Fix finding A in both `sheet-chart.md` copies — escape as `\|` or split into ` / `-joined code spans, and re-check that every row in the tables this PR adds has exactly two rendered columns, not only the AITable ones.
2. Fix finding B — route the projection through `unwrapSheetResult`'s object test, and move `{"result":null}` / `{"result":[]}` out of the rejection list into an accept case that projects into the outer object.
3. For finding C, either restore the three items (an intent-split import row; one `template list/search/apply` route, or a minimal `sheet-template.md`; a compact `dropdown`/`checkbox`/`none` + preserve/clear line) or record each explicitly under `## Notes` as an intentional scope cut, in both the mono and multi trees.

I will re-run V1–V7, both conditional gates, the whole-corpus GFM column count, the `addCreatedSheetID` shape probe across all three revisions, the Python window ablations, and the base-vs-head route counts for import / template / `dataValidation`.

pengzhihan47-star:
@haofeng0705 @maoqxxmm Thanks for the follow-up review. I fixed findings A/B/C on the current head `cd4edf10` (with the routing fix in `e04268d1`).

### Finding A — GFM chart row

- Rewrote the `catAx` / `valAx` row without inline pipe separators in both multi and mono Sheet references.
- The axis contract remains complete: `axisMin`, `axisMax`, `splitLine`, `minorSplitLine`, `axisLabel`, and `axisLine` all render in the second column.
- Re-ran the whole `skills/**/*.md` header/row column-count scan: **12 mismatches**, back to the repository baseline reported in the review (the two PR-introduced mismatches are gone).

### Finding B — non-object `result`

- `addCreatedSheetID` now follows `unwrapSheetResult`: it unwraps only when `result` is a JSON object.
- `result: null`, `[]`, `"ok"`, and `0` now project into the valid outer receipt instead of failing after create/write/readback succeeded.
- The outer business fields are preserved and the probed `nodeId` / `sheetId` are still returned. Object-wrapped responses continue to flatten without leaking the transport envelope or losing large integers.
- Moved `result:null` / `result:[]` out of the rejection cases and added acceptance coverage for all four non-object shapes.

### Finding C — Sheet routing/contracts

- Split local analysis from Excel-to-online import: local xlsx/xls → `sheet import create`; Drive xlsx/xls → `drive download` then import; xlsm/csv remain local because current import does not support them.
- Restored `sheet template list/search/apply` in the creation Golden Route and stage index, including unique `templateId` selection.
- Restored the compact `dataValidation` tri-state: omit = preserve, `type:none` = clear, dropdown/checkbox = write; dropdown `options` and `sourceRange` are mutually exclusive.
- Mirrored only the required Sheet reference changes into mono and added a focused static regression test for import/template/dataValidation/chart routing.

### Verification

- `go test ./internal/helpers -count=1` — PASS (`104.874s`)
- `go test ./test/unit -count=1` — PASS
- focused create-result shape tests — PASS
- `python3 scripts/gen_skill_shortcut_sections.py --check` — PASS (`visible_shortcuts=401 services=20`)
- multi/mono Sheet subtree comparison — PASS
- `git diff --check upstream/main...HEAD` — PASS

The branch includes current `main` and is not behind. Please re-review when convenient.

pengzhihan47-star:
Thanks for the compatibility and workflow-risk review. I addressed the three requested Sheet items in `61fd2827`.

### 1. Backward-compatible create receipt

- Object-wrapped responses now keep the original outer envelope and `requestId`.
- Both the new top-level `nodeId` / `sheetId` and the legacy `.result.nodeId` / `.result.sheetId` paths are present.
- Business fields such as `docUrl` are projected to the top level without removing them from `result`; large JSON integers remain exact.
- Non-object `result` values remain accepted and unchanged.

### 2. Non-atomic create / write / style recovery

There is no backend transaction spanning these remote calls, so the CLI does not pretend to roll them back or auto-delete potentially valid data. Every failure after successful creation now returns a typed, non-retryable `sheet.create_with_data` error with:

- `details.nodeId` and the probed `details.sheetId` when available;
- `details.status` (`unknown` or `partial_success`), `failedStage`, and `retryCreateSafe=false`;
- a stable reason for initialization, rename, write-commit-unknown, verification, style-commit-unknown, or result-projection failure;
- recovery actions that read the existing workbook and only complete missing steps instead of replaying `create-with-data`.

The multi and mono Sheet references carry the same recovery rule and explicitly prohibit auto-deleting written data.

### 3. Readback rollback to `main`

- Removed `verifyRangeNotEmptyWithRetry`, `verifyRangesNotEmptyWithRetry`, the visibility sentinel, and the 250/500/1000ms delays.
- `--values` performs one readback; multi-Sheet mode performs one readback per non-empty Sheet, matching `main`.
- The already-probed default `sheetId` is still reused, so the extra `get_all_sheets` call remains eliminated.

### Verification

- `go test ./internal/helpers -count=1` — PASS (`100.612s`)
- `go test ./test/unit -count=1` — PASS
- `go test ./internal/app -run TestReviewedRoutedInterfacesReachFinalSchema -count=1` — PASS
- `make build` — PASS
- `python3 scripts/gen_skill_shortcut_sections.py --check` — PASS
- `bash scripts/policy/check-command-surface.sh --strict` — PASS
- `bash scripts/policy/check-generated-drift.sh` — PASS, including deterministic registry hashes
- `git diff --check` — PASS

The PR description and release fragment are updated to remove the stale “flat result / bounded readback retry” claims.

liyuan333:
**Report 业务方评审意见**（scope：仅 Report 相关变更，head `efe9a0aa`）

先说结论：`+report-latest` 的分页遍历、跨页去重、时间窗冻结（end=调用时刻）与有界上限（10 页/200 条/2 分钟）的核心逻辑评审通过，是对「>20 篇必报错」真实缺陷的正确修复，fail-closed 契约与本模块既有哲学一致，测试覆盖到位，方向认可 👍。以下 2 点需要处理，1 点建议确认：

**1. report.md 重写后丢失了「按发件人查收件箱」的指引（请补回）**

旧版 `skills/multi/dingtalk-misc/references/report.md` 有明确流程：先 `dws aisearch person --query "<姓名>" --dimension name` 解析 userId，再 `inbox list --sender-user-ids <id>`。新版全文无任何 sender / 发件人相关内容，但 `+inbox-list` 的 `sender-user-ids` flag 仍在（`report.go` ParamDecl 保留声明）。「XX 发给我的日志」是收件箱高频真实意图，Agent 失去这条链路后只能猜测或漏答。请在 Golden Routes 或「收件箱：范围、筛选与分页」章节补一行。

**2. mono 侧 Report 完全未同步（请说明是否有意，或补同步）**

本 PR 重写了 multi 侧 `report_received_today.py`（修掉了「dws 调用失败被静默包装成空结果 + exit 0」的缺陷）并重写了 multi `report.md`，但 mono 侧两个对应文件均保持旧版：

- `skills/mono/scripts/report_received_today.py` — 仍是旧实现（`_internalDetailCommands` 解析 + 失败静默降级），刚在 multi 侧修掉的缺陷在 mono 侧继续存在
- `skills/mono/references/products/report.md` — 仍是旧版全文

同一 PR 里 mono Sheet 已做同步（commit `57ecb17` "fix(skills): sync mono Sheet references"），Report 却没有；且 PR 描述 Scope 写的 "no skills/mono files" 与实际改动（mono Sheet 文件已改）不符。如果 mono Report 是有意分批，请更新 PR 描述并同步修订 `skills/content-qa/mono-multi-coverage.yaml` 中该脚本的 defer 理由（当前 "pending report.md reference" 在本 PR 重写 report.md 后已过时）；否则请补同步。

**3. end=now 的时钟偏差边界（非阻塞，建议冒烟确认）**

时间窗上界从「当天 23:59:59」改为客户端调用时刻，对多页遍历的正确性是必要前提，认可。但 `createTime` 由服务端赋值：若服务端时钟快于客户端，「刚提交完立刻查最新」场景下最新一篇可能落在窗口外。建议用真实账号做一次「`entry submit` 后立即 `+report-latest`」冒烟，确认能命中刚提交的日志。

pengzhihan47-star:
@liyuan333 感谢 Report 业务评审。两项要求已在 46dc016d 修复，并在 c45f323e 同步最新 main。

1. 已恢复按发件人查询收件箱的 Golden Route

- Multi 与 Mono 都明确先执行 dws aisearch person --query "<姓名>" --dimension name --format json。
- 仅在同一 profile 下取得唯一 userId/staffId 后，才执行 dws report +inbox-list ... --sender-user-ids <USER_ID>。
- 零命中、多候选或身份不完整会停止并消歧；禁止默认选第一项或改查他人的发件箱。

2. 已修复 Mono Report 漂移

- skills/mono/scripts/report_received_today.py 已与 Multi 安全实现保持字节一致并恢复可执行位。
- 两侧统一使用 +inbox-list、最多 10 页/200 条/120 秒总时限；删除旧的 _internalDetailCommands 解析、逐条 entry get N+1，以及 dws 失败后返回空结果/exit 0 的行为。
- Mono 保留原有较完整的产品 reference，没有整份替换成压缩版；只补充同等的发件人路由和摘要脚本契约，避免丢失现有业务细节。
- mono-multi-coverage.yaml 现已把两个脚本声明为 identical paired file，并删除过期的 pending report.md reference defer，后续漂移会被 CI 阻断。

新增/更新的回归证据：

- Python helper 回归 8 项全部通过，覆盖 Mono/Multi 一致性、可执行位、失败非空结果、分页去重、边界和超时相关契约。
- TestCrossPlatformCoverageReportSenderRouteAndHelper 通过，锁定两侧人员解析、+inbox-list、sender-user-ids、歧义 fail-closed 和脚本入口。
- go test ./test/unit、go test ./internal/shortcut/report、生成 Skill section 检查和 generated drift 均通过。
- 与最新 main 合并无冲突，且没有 dingtalk-doc 产品改动。

关于非阻塞的服务端/客户端时钟偏差：本次没有在缺少真实复现的情况下扩大时间上界，继续保留调用开始时冻结 end 的分页一致性。真实 entry submit 后立即 +report-latest 的账号冒烟仍应单独执行；若复现，再基于服务端证据引入有界容差，避免未经验证地扩大查询窗口。

haofeng0705 (commented):
## Gate decision

APPROVE-eligible — every gate condition verified at head `9e972fdd`; all blocking findings from the prior review rounds (GFM rendering, timestamps, midnight window, wrapped result shape, import/template/dataValidation routing, mono sync, sender route) are addressed and test-locked on this head.

## Review provenance

- [PASS] Reviewed at PR head `9e972fdd60ac785e5be9aac40f80acc1714d9cf0` in a clean dedicated checkout
- [PASS] Base revision `b19a16003ca3766615f46e18b978b26bf7d2ec0c` (merge-base with `main`)
- [PASS] Head re-checked immediately before submission — same SHA, no drift

## PR description vs repository template

- Template graded against: `.github/pull_request_template.md`
- [PASS] T1 Summary — what (Report/Sheet workflow compression plus correctness fixes) and why (pagination failure on >20 reports, duplicate workbook risk, routing gaps) are both stated
- [PASS] T2 Risk tier — Standard selected; matches the diff (implementation + skill docs, no workflow/policy/package-graph/generated-Schema infrastructure)
- [PASS] T3 Release fragment — `.changes/1136-report-sheet-workflow-correctness.md` added
- [PASS] T4 Release-seal validation — N/A with valid reason (ordinary fragment PR, `CHANGELOG.md` untouched)
- [PASS] T5 Targeted test/check commands and results — 10 commands listed with outcomes; all reproduced below (V1–V10). Minor staleness: description says 7 Python tests, the current head runs 8 (V6) — observation, not a template defect
- [PASS] T6 Behavior evidence — named tests map 1:1 to changed behavior (pagination, dedup, envelope preservation, single readback, midnight window, timezone)
- [PASS] T7 Conditional checks — documentation rendering checked (GFM scan, see below); full local suite correctly N/A for Standard tier; `check-generated-drift.sh` and `check-command-surface.sh --strict` both run and reproduced; package-manager check N/A (no packaging surface)
- [PASS] T8 Notes — recovery semantics, readback rollback to `main`, 10-page/200-item fail-closed scope cut, and restored success-critical limits are all documented
- Observations outside the template: none counted as defects

## Author validation reproduction

| ID | Acceptance criterion | Author claim / exact step | Documented environment | Independent result | Match | Evidence |
|---|---|---|---|---|---|---|
| V1 | Build succeeds | `make build` — PASS | local | PASS | Yes | exit 0 |
| V2 | Helpers suite green | `go test ./internal/helpers -count=1` — PASS | local | PASS | Yes | `ok ... 98.509s`, exit 0 |
| V3 | Unit docs tests green | `go test ./test/unit -count=1` — PASS | local | PASS | Yes | `ok ... 1.030s`, exit 0 |
| V4 | Report package green | `go test ./internal/shortcut/report -count=1` — PASS | local | PASS | Yes | `ok ... 0.706s`, exit 0 |
| V5 | Final-Schema contract test | `go test ./internal/app -run TestReviewedRoutedInterfacesReachFinalSchema -count=1` — PASS | local | PASS | Yes | `ok ... 1.095s`, exit 0 |
| V6 | Python helper regressions | `python3 scripts/dev/report_received_today_test.py` — PASS | local | PASS | Yes | `Ran 8 tests ... OK` (description says 7; count grew with the dedup test — actual result is stronger) |
| V7 | Skill sections generated | `python3 scripts/gen_skill_shortcut_sections.py --check` — PASS | local | PASS | Yes | `visible_shortcuts=401 services=20`, exit 0 |
| V8 | Command surface intact | `bash scripts/policy/check-command-surface.sh --strict` — PASS | local | PASS | Yes | `command surface check: ok`, exit 0 |
| V9 | No generated drift | `bash scripts/policy/check-generated-drift.sh` — PASS incl. identical hashes | local | PASS | Yes | two assemblies `registry_hash=sha256:ffbe3f7b…` identical; `generated drift check: ok` |
| V10 | No whitespace damage | `git diff --check` — PASS | local | PASS | Yes | exit 0 |
| V11 | Mono/Multi consistency (from follow-up comments) | scripts byte-identical; sheet refs mirror apart from links | local | PASS | Yes | `cmp` identical (11301 bytes, exec bit both sides); link-normalized diff of `sheet-write-data.md` = 0 lines; `check-mono-multi-skill-content.sh` exit 0 |
| V12 | GFM column count back to baseline (from follow-up comments) | 12 mismatches = repository baseline | local | PASS | Yes | full `skills/**/*.md` scan finds exactly 12 mismatches; none on a line this PR adds or rewrites (the one mismatch in a PR-touched file, `aitable/SKILL.md:57`, is byte-identical at base and head) |
| V13 | New Sheet coverage tests | named in behavior evidence | local | PASS | Yes | all 12 new/renamed Sheet tests PASS |

## Behavioral and regression checks

- [PASS] Report latest bounded pagination — `TestCrossPlatformCoverageReportLatestRequiresCompleteOrderedListAndExactReadback`: two-page walk with cursors `0 → 20`, then exact readback of `report-2`; fail-closed branches (conflicting duplicate, stalled cursor, >200 items, 10-page exhaustion) each reproduced with the expected error and call count
- [PASS] Frozen end-of-window — `TestCrossPlatformCoverageReportLatestExecutorRemainingBranches` asserts `endTime` lands inside the invocation snapshot and the range is exactly `reportMaximumOutboxDays`; caller observed a bounded deadline on every call
- [PASS] Sheet probed `sheetId` reuse — `TestSheetCreateReusesProbeSheetID` / `TestCrossPlatformCoverageSheetCreateReturnsProbedSheetID`: no extra `get_all_sheets` after probe
- [PASS] Backward-compatible receipt — `TestCrossPlatformCoverageAddCreatedSheetIDPreservesLegacyEnvelopeAndProjectsTopLevel`: top-level `nodeId`/`sheetId` added, `.result.nodeId` retained, outer `requestId` kept, large integers exact; non-object `result` shapes accepted (`...KeepsOuterObjectForNonObjectResult`)
- [PASS] Single readback aligned with `main` — `TestCrossPlatformCoverageSheetCreateReadbackMatchesMainSingleAttempt`; code inspection confirms `verifyRangeNotEmpty` is one call with no 250/500/1000ms retry loop
- [PASS] Structured non-retryable recovery — `TestCrossPlatformCoverageSheetCreateWriteFailureIsNotBlindlyRetryable`: `retryable=false`, `retryCreateSafe=false`, stage/status/reason details; projection failure also covered
- [PASS] Context handling — cancelled-context tests for `waitSheetWritable` / retry delay; command context propagated (`TestSheetCreatePropagatesCommandContext`)
- [PASS] Midnight window + readable timestamps — `test_query_window_clamps_first_second_after_midnight`, `test_format_create_time_uses_shanghai_timezone` (`1970-01-01 08:00:00 +0800`)
- [PASS] Sender route + mono sync — `TestCrossPlatformCoverageReportSenderRouteAndHelper` pins `aisearch person` → `+inbox-list --sender-user-ids` with fail-closed ambiguity on both sides; helper scripts byte-identical
- [PASS] Prior-round routing findings — import vs local-analysis split, `sheet template list/search/apply`, and the `dataValidation` tri-state are present in both mono and multi references and locked by `TestCrossPlatformCoverageSheetSkillRoutesImportTemplatesAndValidation`
- [PASS] Restored success-critical limits — Report `--size <= 20` and 10MB content cap in `report.md`; chart `title`/`legend`/`catAx`/`valAx` field set in `sheet-chart.md` (axis row rewritten without inline pipes); `--styles` unknown-key rejection retained at both levels

## Release fragment validation

- [PASS] F1 Fragment present and legal — `.changes/1136-report-sheet-workflow-correctness.md`, flat, regular file, name matches `^[a-z0-9][a-z0-9._-]*\.md$`
- [PASS] F2 Category valid — `Changed`
- [PASS] F3 Body well-formed — four Markdown list items, no TODO/TBD
- [PASS] F4 `CHANGELOG.md` untouched — zero changed entries in the base...HEAD diff
- [PASS] F5 English-only, no rival names — CJK scan: 0 hits; rival scan: 0 hits on PR-added fragment lines
- [PASS] F6 Fragment matches the diff — each of the four bullets verified against `latest.go`, `sheet_create_with_data.go`, the routing references, and the helper scripts; no overstatement found
- [PASS] `check-release-fragments.sh b19a1600 9e972fdd` — exit 0
- Baseline: no pre-existing non-English or rival-name issues in scope; repo-wide rival scan over all PR-added lines: 0 hits

## Conventional code review

- [PASS] Diff scope and surrounding call paths — 40 files read; Go changes confined to `sheet_create_with_data.go` and `report/latest.go` plus tests; no unrelated edits
- [PASS] Error handling and failure paths — every post-create failure stage returns a typed `sheet.create_with_data` error with `nodeId`/`sheetId`/stage/status; projection failure fails closed instead of printing partial output
- [PASS] Concurrency, cancellation, cleanup, lifecycle — probe retries now honor `ctx` via select (previously ignored); `+report-latest` wraps the whole walk in a 2-minute timeout and restores the original command context in `defer`
- [PASS] Security, permissions, temporary files, command execution — no new exec/file/secret surface; `RawMessage` projection avoids float64 round-tripping of large integers
- [PASS] Backward compatibility and configuration defaults — legacy `.result.nodeId` and outer `requestId` retained; readback behavior restored to `main` parity; release-test hardening (`maintenance.auto=false`, `gc.auto=0`) is a justified TempDir-race fix
- [PASS] Platform-specific files and build tags — coverage tests carry the `TestCrossPlatformCoverage*` prefix required by the platform gate selector
- [PASS] Test coverage quality — every changed branch has a focused test; item/page-limit, overlap-dedup, conflicting-duplicate, and cancellation paths all covered
- [PASS] Formatting, static analysis, generated files, dependency scope — drift gate green with identical registry hashes; no new dependencies; one cosmetic observation: `test/unit/sheet_skill_docs_test.go` omits the Apache header its sibling files carry
- [PASS] Documentation consistency with the code — routing references, fragment, and Long/help prose describe the same behavior as the code

## CI and baseline comparison

- [PASS] Required CI checks on head `9e972fdd` — 36 success, 7 skipped, 0 failure (including Test matrix, race shards, Policy-relevant gates, AI Behavior); PR is not a draft, so this is the full gate set
- [PASS] Claimed known failures reproduced on PR — none claimed
- [PASS] Base comparison — no mismatch required one; all V1–V13 reproduced the author's conclusions directly

## Findings and blockers

- Blocking code findings: none
- Description blockers (template-required only): none (V6 count staleness is an observation)
- Fragment blockers: none
- Validation mismatches: none
- Not-verified items: real-account clock-skew smoke after `entry submit` (reviewer liyuan333's non-blocking item) — deliberately deferred by the author until reproduced with server-side evidence; accepted as stated

## Re-review condition

No changes required for the gate. Optional, non-blocking: add the Apache license header to `test/unit/sheet_skill_docs_test.go` and refresh the Python test count in the description (7 → 8). Approval itself is pending maintainer confirmation; the two earlier change-requests on this PR were submitted against stale SHAs, and every blocking finding they listed is addressed on this head.

pengzhihan47-star:
感谢完整复评并确认当前版本已达到 APPROVE-eligible，所有既有阻塞项均已关闭。复评后 main 继续前进，PR 与 main 在 Mono/Multi QA 清单产生单文件冲突；我已在 6731110f 完成同步并按双方契约合并：保留 Report 脚本配对，同时保留 main 新增的 Minutes 脚本配对和退役清单，未恢复过期的 Report orphan 例外。验证结果：Mono/Multi 内容检查通过，Skill section 生成检查通过，test/unit 通过，diff check 通过，Schema compatibility 对最新 main 和 v1.0.59 均通过。相对最新 main 的 PR 差异仍限于既有 Report、Sheet、AITable 范围，没有 dingtalk-doc 产品改动。请在新 head CI 完成后确认正式 approval。可选的 Apache header 和描述中的 Python 测试数量更新均不影响本轮 gate，本次未为其增加额外代码改动。

haofeng0705 (approved):

haofeng0705 (approved):
Review SHA = `d81df158`, evidence SHA = `9e972fdd` — the two head moves since the evidence review were merge-only (upstream/main syncs bringing minutes and doc/drive import work). All implementation files of this PR are hash-identical between the evidence SHA and this head (verified per file: `sheet_create_with_data.go`, `report/latest.go`, the fragment, both helper scripts, both docs tests, the Sheet/report references); the only differing files (`gen_skill_shortcut_sections.py`, `mono-multi-coverage.yaml`) differ by the additive merged minutes entries, with this PR's report/sheet entries intact. All evidence conclusions therefore stand as published in the preceding review comment.

Approving on the gate result: description template T1–T8 satisfied; all 13 validation items independently reproduced PASS at the evidence SHA; release fragment F1–F6 clean (English-only, no rival names, matches the diff); no blocking code-review findings; CI fully green on this head. Two optional, non-blocking nits noted in the evidence comment: license header on `test/unit/sheet_skill_docs_test.go` and the stale Python test count in the description (7 → 8).

PR #1285, opened 2026-09-04
to #dws apply fix

hlzjsong:
## Summary

Refactors the CLI auth-apply OAuth frontend: moves the "apply pending" state from the inline success-msg component embedded in `notEnabledHTML` into a dedicated `applyPendingHTML` page, and strengthens idempotency handling for apply requests.

### Frontend
- **New `applyPendingHTML` page**: a standalone "access permission request pending" page with a clock icon and polling logic; auto-redirects to the success page once the admin approves
- **Simplified `notEnabledHTML` page**: removes the inline success-msg component and polling logic; a successful apply now redirects via `location.href="/applyPending"`
- **Double-click protection**: new `applying` flag prevents duplicate submissions
- **`DWS_USE_APPLY_DUPLICATE` handling**: the frontend recognizes the duplicate-apply error code and lands on the pending page instead of showing a retryable error
- **`accessDeniedHTML` copy update**: unified wording to "该组织尚未开启CLI数据访问权限"

### Backend
- **Three-way routing**: the `/callback` handler now decides between success / applyPending / notEnabled based on `cliAuthEnabled`, `applySent`, and `hasDwsApply`
- **New `/applyPending` endpoint**
- **`Cache-Control: no-store`**: applied globally to prevent stale cached pages or `/api/status` responses from re-showing the apply form after the server reports `hasDwsApply`
- **`hasDwsApply` flag**: recognizes an existing apply request from a previous session, avoiding the form being shown again after a page refresh
- **Apply idempotency**: `isAlreadyAppliedError()` normalizes `DWS_USE_APPLY_DUPLICATE` to success so the retry layer does not report a spurious failure
- **Terminal polling enhancement**: the polling loop additionally checks `hasDwsApply` to support cross-session pending detection
- **`CLIAuthResult.HasDwsApply` field**: surfaces the server-side existing-apply status

### Tests
- `oauth_provider_coverage_test.go`: covers three-way routing, `hasDwsApply` redirect, and idempotency scenarios
- `auth_extra_test.go`: unit tests for `isAlreadyAppliedError`

## Files Changed
- `internal/auth/oauth_helpers.go` — added `applyPendingHTML`, simplified `notEnabledHTML`, added `isAlreadyAppliedError` and `CLIAuthResult.HasDwsApply`
- `internal/auth/oauth_provider.go` — three-way routing, `/applyPending` endpoint, `no-store` header, idempotency handling, polling enhancement
- `internal/auth/oauth_provider_coverage_test.go` — new tests
[...]

PR #1331, opened 2026-09-08
fix(mcp): restore mcpdev direct-runtime endpoint pin

typefield:
## Summary

- Restore the `mcpdev` product pin in `internal/app/direct_runtime.go` that merge `d92a0bdd` dropped, so `dws dev mcp` resolves an endpoint again
- Derive the endpoint from `defaultPATGatewayBaseURL()` (the `devapp` pattern) instead of the pre-migration hardcoded pre-release gateway, so it follows the active environment
- Add `internal/app/direct_runtime_test.go` and extend the existing `TestCrossPlatformCoverageDirectRuntimeCoverage` gate so a future merge that drops the pin fails CI

## Root cause

`mcpdev` is a **helper-only** product. It is absent from the generated static
server table (`internal/syncdata/endpoints.go`, `routing.go` — both
`Code generated by cmd/sync-oss. DO NOT EDIT.`) and it never appears in MCP
discovery, so the source pin in `direct_runtime.go` is its **only** default
resolution path.

Merge `d92a0bdd` resolved a conflict in the same `const` block where `main` had
added `recruitProductID`, and took the `main` side wholesale:

```
-   mcpdevProductID       = "mcpdev"
-   mcpdevEndpointEnv     = "DINGTALK_MCPDEV_MCP_URL"
-   preMCPGatewayBaseURL  = "https://pre-mcp-gw.dingtalk.com"
-   mcpdevServerPath      = "/server/62445d67..."
+   recruitProductID      = "recruit"
```

`mcpdevMCPEndpoint()` and both of its tests disappeared with it. Meanwhile
`internal/helpers/dev.go` still mounts the `dws dev mcp` tree unconditionally and
`dev_mcp_contract.go` still declares `Availability: contract.InterfaceAvailable`.

Result: all three resolution paths miss (static table, hardcoded built-in,
dynamic registry) and every `dws dev mcp` subcommand fails with
`endpoint_not_resolved` — raised in `handleCatalogMiss` **before**
`executeInvocation`, so no HTTP request is ever sent. This is a regression, not
a missing feature: the pin existed and worked from `f98c09b0` until `d92a0bdd`.

## Note for reviewers

The pre-migration implementation hardcoded `pre-mcp-gw.dingtalk.com`. This PR
deliberately does **not** restore that: it uses `defaultPATGatewayBaseURL()`, so
`mcpdev` now defaults to production and follows `~/.dws/mcp_url`, exactly like
[...]

PR #1334, opened 2026-09-08
feat: refresh runtime context and login URLs

zearlin:
## Summary

Refresh the embedded runtime payload to `20260908`. DWS now verifies and publishes owned resources beside the resolved executable, falls back to its versioned cache on conflicts or publication failures, and attaches the immutable runtime context to OAuth and Device Flow browser URLs. Terminal links and diagnostics remain redacted.

Adjacent publication uses a cross-process lock and pending/ready ownership records. A warm reuse reads the trusted manifest directly from the verified embedded archive and hashes each published resource once, without staging or extraction. Only installation and repair materialize a new copy. OAuth reauthorization consumes the server-generated complete URL; Device Flow reuses one snapshot across retries. Release, npm and Homebrew still distribute one binary. Workflow files are identical to the synchronized `main` baseline.

## Risk tier

- [ ] Documentation-only
- [ ] Standard
- [x] High-risk: auth, platform payloads, recovery, installers and packaging

## Verification

- [x] Release fragment: `.changes/runtime-context-refresh.md`
- Release-seal validation: N/A; `CHANGELOG.md` is unchanged.
- [x] Targeted runtimepayload, runtimecontext and auth package tests.
- [x] Race tests for adjacent publication, process locking, URL handling and immutable login snapshots.
- [x] `go test ./internal/runtimepayload ./internal/runtimecontext -run '^TestCrossPlatformCoverage' -count=1 -coverprofile=/tmp/pr1334-platform-coverage.out`: both packages have 100% statement coverage, including the warm reuse fix.
- [x] `go test -race ./internal/runtimepayload ./internal/runtimecontext -count=1`: passed.
- [x] `go test ./internal/runtimepayload -run '^$' -bench '^BenchmarkRuntimePayloadWarm$' -benchmem -benchtime=7x -count=3`: real embedded payload on darwin/arm64 (Apple M4 Pro). Adjacent warm reuse fell from 108.2–164.6 ms/op to 14.0–17.2 ms/op; cache reference remains 12.4–13.0 ms/op. Allocations fell from about 14.61 MB / 19,686 allocs per operation to 4.33 MB / 2,650 allocs. The benchmark never initializes the native SDK.
- [x] `TestCrossPlatformCoverageAdjacentEmbeddedWarmReuse` rejects staging/extraction on reuse and asserts one checksum pass across all 124 real resources. `TestCrossPlatformCoverageAdjacentRejectsSelfAssertedChecksums` proves a forged on-disk checksum is repaired against the embedded manifest.
- [x] Five library architectures, eight exports, 128 checksums and deterministic six-target embedded assets.
- [x] Six `CGO_ENABLED=0` cross-builds; `make build`.
- [x] `make policy`, including generated drift and Schema gates.
- [x] Finalized release archives and isolated npm/Homebrew package-manager verification.
- [x] Native macOS doctor reports ready, payload `20260908`, nonzero token length and a 12-character fingerprint. Symlink launch materializes into real `libexec`; working and symlink directories remain untouched.
- [x] Original implementation full suite: `DWS_PACKAGE_VERSION=0.0.0-test go test -p 2 ./...` passed locally. The initial concurrent run had three unrelated plugin/Gitee fixture failures; all passed isolated reruns and the complete rerun. The reviewer reproduced an environment-dependent SIGKILL in the Schema subprocess on both base and head under memory pressure, so this is not a guarantee for all host/concurrency conditions.
- Follow-up integration validation: the constrained `GOMEMLIMIT=1GiB GOGC=50 ... go test -p 2 ./...` run hit the app package 10-minute timeout and a Schema subprocess SIGKILL. It was stopped before the next main synchronization. The replacement command is `GOMEMLIMIT=8GiB DWS_PACKAGE_VERSION=0.0.0-test go test -p 1 -timeout 30m ./...`; the latest-head Actions run independently validates the integrated branch. Neither the interrupted run nor a cross-build is counted as a passing native execution test.
- Documentation: `docs/runtime-context.md` describes the final behavior and flow.

## Notes

Local packaging uses ad-hoc Darwin signing. Official Apple Developer ID signing and native execution on Linux/Windows remain release/CI validations; cross-compilation does not establish those results. Native doctor validation establishes a ready context, but does not establish prompt-free macOS Keychain access: the embedded SDK may request system authorization. No workflow, job or step was changed. Main synchronization also includes the approval-pending login page from PR #1285; reauthorization continues to consume the complete server-generated URL and Device Flow retains one runtime snapshot across retries. Check the latest-head Actions run and review decision for merge readiness.


haofeng0705 (changes requested):
## Gate decision

REQUEST CHANGES — every validation claim reproduced except one baseline/environmental item, but the new adjacent-publication warm path adds ~117 ms to essentially every real invocation of the CLI, and that cost is avoidable.

## Review provenance

- [PASS] Reviewed at PR head `cbc17657a8fe5275d40f830a54d3d3bf2dbff58e` in a clean dedicated checkout
- [PASS] Base revision `dba9f061049f3799983220c3c59abfeca8207306` — current `origin/main`, and `merge-base HEAD origin/main` equals it, so the branch sits directly on top of main
- [PASS] Head re-checked before submission — `cbc17657`, no drift; every CI check run also carries `head_sha = cbc17657`

## PR description vs repository template

- Template graded against: `.github/pull_request_template.md`
- [PASS] T1 Summary — states what changed ("Refresh the embedded runtime payload to `20260908`… verifies and publishes owned resources beside the resolved executable, falls back to its versioned cache on conflicts or publication failures, and attaches the immutable runtime context to OAuth and Device Flow browser URLs") and the operational rationale (cross-process lock, pending/ready ownership records, one snapshot across retries, "Release, npm and Homebrew still distribute one binary").
- [PASS] T2 Risk tier — exactly one tier selected: `[x] High-risk: auth, platform payloads, recovery, installers and packaging`. Matches the diff, which touches `internal/auth`, `internal/runtimepayload/assets/*`, `scripts/install.ps1` and `scripts/release/post-goreleaser.sh`.
- [PASS] T3 Release fragment — `.changes/runtime-context-refresh.md` added; `CHANGELOG.md` untouched.
- [PASS] T4 Release-seal validation — claimed `N/A; CHANGELOG.md is unchanged`. The reason holds; this is not a release-seal PR. `check-changelog-pr.sh --fast-path` / `--content-only` are therefore N/A too, not a defect.
- [PASS] T5 Targeted test/check commands and results — exact commands with results for the full suite (`DWS_PACKAGE_VERSION=0.0.0-test go test -p 2 ./...`), `make build`, `make policy` and six `CGO_ENABLED=0` cross-builds. The quantified results are precise and independently reproducible ("100% coverage (156 statements)", "Five library architectures, eight exports, 128 checksums", "deterministic six-target embedded assets"). Observation, not a defect: the two targeted-test items name packages and test categories rather than literal invocations.
- [PASS] T6 Behavior evidence — concrete CLI output shape: "Native macOS doctor reports ready, payload `20260908`, nonzero token length and a 12-character fingerprint. Symlink launch materializes into real `libexec`; working and symlink directories remain untouched." Reproduced exactly (V9).
- [PASS] T7 Conditional checks — full local suite recorded because the tier is high-risk; generated drift and Schema gates covered by "`make policy`, including generated drift and Schema gates"; packaging covered by "Finalized release archives and isolated npm/Homebrew package-manager verification". Documentation rendering is N/A (not documentation-only), though the added `docs/runtime-context.md` is still described. `check-command-surface.sh --strict` is N/A — no command or flag surface changed.
- [PASS] T8 Notes — discloses "Local packaging uses ad-hoc Darwin signing", that official Apple Developer ID signing and native Linux/Windows execution remain release/CI validations, and that "No workflow, job or step was changed". The last claim is verified: the diff against `.github/` is empty.
- Observations outside the template: none counted as defects. Not requested by this template, so not raised: local OS/arch/Go version, a requirement-to-acceptance-criterion mapping table, a separate limitations section, or reproducing the whole CI suite locally.

## Author validation reproduction

| ID | Acceptance criterion | Author claim / exact step | Independent result | Match | Evidence |
|---|---|---|---|---|---|
| V1 | Fragment exists for the user-visible change | `.changes/runtime-context-refresh.md` | PASS | Yes | Regular file, flat at the top level of `.changes/`, name matches `^[a-z0-9][a-z0-9._-]*\.md$` |
| V2 | Targeted package tests pass | "Targeted runtimepayload, runtimecontext and auth package tests" | PASS | Yes | `go test ./internal/runtimepayload/... ./internal/runtimecontext/... ./internal/auth/...` → `ok` 5.872s / 2.747s / 134.739s (Go 1.26.5, darwin/arm64) |
| V3 | Race-clean adjacent publication, process locking, URL handling, immutable login snapshots | "Race tests for adjacent publication, process locking, URL handling and immutable login snapshots" | PASS | Yes | `go test -race -count=1 ./internal/runtimepayload/... ./internal/runtimecontext/... ./internal/event/lock/...` → all `ok` (9.642s / 5.346s / 2.382s). `go test -race ./internal/auth/ -run 'TestCrossPlatformCoverageOAuthRuntimeURLs\|TestCrossPlatformCoverageDeviceRuntimeURLRetrySnapshot' -v` → both PASS across all state subtests |
| V4 | Changed executable statements fully covered | "Changed executable statements: 100% coverage (156 statements)" | PASS | Yes | Reproduced **exactly**: intersecting added HEAD line numbers with merged uncovered coverage blocks gives **156 changed executable statements, 0 uncovered**. `adjacent.go` 3/3 funcs, `runtimecontext/context.go` 14/14 (incl. new `AttachToURL`), `payload.go` all 100%. The sub-100% functions in `oauth_provider.go` (`GetTokenSnapshot` 95.5%, `persistKnownLoginToken` 70.0%) and `device_flow.go` (`loginOnce` 97.8%) sit on **untouched** lines — pre-existing gaps, not PR-introduced |
| V5 | Payload integrity and asset determinism | "Five library architectures, eight exports, 128 checksums and deterministic six-target embedded assets" | PASS | Yes | `./scripts/policy/check-runtime-payload.sh --allow-unsupported-tools` → exit 0, `Runtime payload verified: 5 libraries, 6 targets, 123 ps files.` and **no** "Skipping … export inspection" lines, so all eight symbols were really inspected. `./scripts/build/generate-runtime-payload-assets.sh --check` → exit 0, `Embedded runtime payload assets are current.` Independently, `shasum -a 256` on all five libraries matches both `manifest.json` and `SHA256SUMS`; 128 = 5 libraries + 123 `ps` files, agreeing with `ps_file_count: 123` |
| V6 | Cross-compilation and local build | "Six `CGO_ENABLED=0` cross-builds; `make build`" | PASS | Yes | All six OK — darwin/amd64 53970288, darwin/arm64 51407442, linux/amd64 53027416, linux/arm64 49968538, windows/amd64 60035584, windows/arm64 52432384 bytes. `make build` exit 0 → `dws` 40484512 bytes, payload attached and signed |
| V7 | Policy gates pass | "`make policy`, including generated drift and Schema gates" | PASS (CI on this SHA + the two added gates locally) | Yes | The CI `Policy` job runs literally `make policy` (`ci.yml:2811`) and is green on `cbc17657` in 11m25s. Locally reproduced the two gates this PR *adds* to `policy` — `check-runtime-payload.sh` and `generate-runtime-payload-assets.sh --check`, both exit 0 — plus `check-release-fragments.sh` exit 0 |
| V8 | Packaging and installer surfaces verified | "Finalized release archives and isolated npm/Homebrew package-manager verification" | NOT VERIFIED | n/a | `scripts/release/verify-package-managers.sh` needs `dist/homebrew/dingtalk-workspace-cli-local.rb` and `dist/npm/…` from `make package` (6-platform `build-all.sh` + `post-goreleaser.sh` + signing), then performs real npm and Homebrew installs — not reproducible in this review environment. Partial mitigation: CI `Test (workflow and release contracts)` and `Validate Runtime Payload` are green on this SHA, and this PR's packaging-side changes are limited to mechanical `20260825` → `20260908` string bumps that I verified are complete |
| V9 | Native macOS runtime context works end to end | "Native macOS doctor reports ready, payload `20260908`, nonzero token length and a 12-character fingerprint. Symlink launch materializes into real `libexec`; working and symlink directories remain untouched." | PASS | Yes | `dws doctor --json` → `runtime_context` `status=pass`, `state=ready`, `payload_version='20260908'`, `token_length=88` (nonzero), `token_fingerprint='e151adfa1889'` (**exactly 12 chars**), `environment='online'`; the token itself appears nowhere in the output. Symlink test with a real `bin/dws -> libexec/dws` launched from an unrelated cwd → resources landed **only** in `libexec/` (`.dws-runtime-manifest.json`, `.dws-runtime.lock`, `ps`, `x7k2m9p4q1w8.dylib`); `bin/` still held just `dws` and the cwd stayed empty |
| V10 | Full local suite passes | "Full suite: `DWS_PACKAGE_VERSION=0.0.0-test go test -p 2 ./...` passed. The initial concurrent run had three unrelated plugin/Gitee fixture failures; all passed isolated reruns and the complete rerun." | MISMATCH — classified **flaky / environment-dependent** (memory pressure); base fails the same check | No (claim is unqualified, not false) | Exact command run twice → both times 132 `ok` and one failure, `internal/cli`: `schema package-cli TestMain: run cmd_schema_catalog: signal: killed` (16.086s / 14.736s). Isolated rerun `go test -count=1 ./internal/cli/` → **`ok` 135.603s**. **Base fails the same pairing that head passes** — see the baseline comparison below. The three plugin/Gitee failures the author mentions did not reproduce in either run |
| V11 | Documentation matches shipped behavior | "`docs/runtime-context.md` describes the final behavior and flow" | PASS | Yes | Spot-checked the falsifiable claims: "123 data files" (`ls ps` = 123), "Five libraries" (5), "Win32 is not supported" (`LibraryName` has no `386` case), "Only `k9Xm2pQv` is called; the other seven exports are checked for integrity but are not invoked" (2 refs to `k9Xm2pQv` via `initSymbol` in `native_unix.go`/`native_windows.go`; **0** refs to the other seven), plus the symlink-resolution and cache-path descriptions (both reproduced in V9) |

## Behavioral and regression checks

- [PASS] Adjacent publication lands beside the **resolved** executable — verified natively (V9) and covered by `TestCrossPlatformCoverageAdjacentActualSymlinkAndWorkingDirectory`, which uses a real symlink plus real `filepath.EvalSymlinks` and asserts nothing is written to the symlink directory or the cwd.
- [PASS] Unknown resources are never clobbered — `TestCrossPlatformCoverageAdjacentProtectsUnknownResources` pre-plants user files/dirs at `libx7k2m9p4q1w8.so`, `ps`, `.dws-runtime-manifest.json` and `.dws-runtime.lock`, asserts rejection *and* that the user bytes are unchanged; the `owned symlink` subtest asserts a planted symlink is rejected and left in place. I traced the same guarantee in the source: `readOwnership` fails closed when the record is unreadable, so a second Unix user sharing an install directory falls back to cache instead of fighting over the first user's files.
- [PASS] Interrupted publication recovers — `TestCrossPlatformCoverageAdjacentInterruptedPublication` injects a rename failure at each of the four rename points and asserts the next invocation commits `ready`. Error text stays neutral (asserts `sensitive`/`private` never appear).
- [PASS] Cross-process lock is genuinely cross-process — `TestCrossPlatformCoverageAdjacentProcessLock` re-execs the test binary as a real child holding the lock, asserts the parent is rejected, then asserts success after release. Backed by `lock.TryAcquire` = `flock(LOCK_EX|LOCK_NB)` with an `ErrBusy` sentinel; `MaterializeAdjacent` treats *any* lock error as fail-closed to cache, which is the safe direction.
- [PASS] Cache-fallback matrix — `resolveLibraryPath` covered for adjacent failure, `os.Executable` failure, `EvalSymlinks` failure, both-fail and no-cache-directory, asserting the neutral `runtime payload unavailable` message.
- [PASS] URL attachment contract — `TestCrossPlatformCoverageAttachToURL` uses an adversarial token (`private+runtime/&=测试 value`) and a URL carrying duplicate `callerUmt`/`caller` params plus a fragment; asserts single-valued replacement, other params and fragment preserved, and idempotency. The fail-open set is well chosen: `%`, `?bad=%zz`, `?a=1;b=2` (semicolon), `file:///`, `/relative`, `https://user:secret@…` (userinfo rejected), plus zero/Timeout/Error/empty-token/control-char contexts.
- [PASS] Login snapshot immutability — one `Resolve()` per login asserted for both flows (`resolutions != 1` fails); Device Flow reuses one snapshot across all three attempts (`opens != 3` fails) and asserts the server response is unmutated (`got.VerificationURIComplete != raw` fails).
- [PASS] Redaction — terminal and manual links keep original URLs; browser-launch errors now log a neutral `error_category` instead of the launcher's error text, which could have contained the URL. `TestCrossPlatformCoverageDoctorRuntimeContextIsRedacted` asserts the secret appears in neither the rendered output nor the JSON-encoded detail, and the dry-run test asserts the context is resolved **zero** times.
- [PASS] OAuth page simplification is behavior-preserving — deleting the client-side URL construction in `notEnabledHTML` is safe: `buildAuthURLForRegion` already supplies `prompt=consent`, `response_type=code`, `scope=openid corpid`, `client_id` and `redirect_uri`, and `redirectURI` is built from the *actually bound* port (`listener.Addr().(*net.TCPAddr).Port`), matching the old `location.port`. It also now carries `corpId` and `lang`, which the old hand-built `backLink` dropped — an improvement that makes the reauthorization link consistent with the initially opened browser URL.
- [PASS] The removed sidecar fallback is dead code, not a compatibility break — the repo's own release verification forbids that layout: `verify-release-artifacts.sh:110` fails an archive containing `.dws-runtime` ("contains a legacy sidecar runtime payload"), and `verify-package-managers.sh:303`/`:374` fail npm and Homebrew installs that retain one. `post-goreleaser.sh` attaches the payload into the binary and then removes the staged sidecar. One binary is the shipped contract, so the deleted discovery branch could not fire for a supported install.
- [PASS] Version bump is complete — `git grep 20260825` over the HEAD tree returns only `docs/drafts/oa-three-role-case-evaluation-20260828.md` (unrelated zip filenames, pre-existing) and `internal/runtimepayload/adjacent_test.go:64`, which *intentionally* sets a stale `PayloadVersion` to exercise the version-mismatch repair path. Both superseded library digests (`7a19607b…`, `1faf5213…`) return zero hits. `20260908` appears in exactly the 14 expected files.
- [PASS] `.gitignore` additions work as intended — after running the built `dws` inside the checkout, `git status --porcelain` stayed empty while `.dws-runtime-manifest.json`, `.dws-runtime.lock`, `ps/` and `x7k2m9p4q1w8.dylib` were present on disk; `git status --ignored=matching` confirms they are ignored rather than missing.
- [FAIL] **Warm-path cost of adjacent publication** — blocking finding below.

## Release fragment validation

- [PASS] F1 Fragment present and legal — `.changes/runtime-context-refresh.md`; flat at the top level of `.changes/`, regular file (not a symlink), name matches `^[a-z0-9][a-z0-9._-]*\.md$`. The change is user-visible (payload refresh, new materialization behavior, login URLs), so a fragment is required and `N/A` would not hold.
- [PASS] F2 Category valid — `category: Changed`, one of the six permitted values, and the right one: the runtime context feature already existed, this refreshes and extends it.
- [PASS] F3 Body well-formed — one Markdown list item, no `TODO`/`TBD`.
- [PASS] F4 `CHANGELOG.md` untouched — confirmed; not a release-seal PR.
- [PASS] F5 English-only, no rival names — **0 PR-introduced hits**. Both scans over the fragment's added lines are empty, and the new user-facing `docs/runtime-context.md` is clean on both scans too. The detector was sanity-checked before being trusted: the word-bounded alternation matches `lark-cli` and `Lark suite` while `Clarkson` / `larkspur` / `clark` produce no match.
- [PASS] F6 Fragment matches the diff — each claim maps to code: payload `20260908` (`payload.go:27` plus the renamed `third_party/runtimepayload/20260908/`), "verify and materialize owned resources beside the resolved executable" (`adjacent.go` + `resolveLibraryPath`), "with cache fallback" (the `materializeRuntimePayload` branch), "attach the process context to browser login URLs" (`AttachToURL` + both login flows), "keeping terminal links and diagnostics redacted" (`error_category` logging, `DiagnosticDetail`). Neither overstates nor omits; no breaking-change callout is needed since nothing user-invocable changed shape.
- [PASS] `check-release-fragments.sh` — `./scripts/policy/check-release-fragments.sh dba9f061 HEAD` → exit 0, no output.
- Baseline: pre-existing CJK on files this PR does **not** touch is `.changes/README.md` (22) and `.changes/schema-required-group-relaxation.md` (1) — context only, not a defect. This PR adds 0.
- Observation, not a defect: the fragment body omits a `(#1334)` reference and the filename is not PR-number-first. Both are optional — `.changes/README.md` says the name is "通常是 PR 号" ("usually"), and existing fragments (`chat-active-conversations.md`, `standalone-whiteboard.md`, `contact-invite-apply-admin.md`) use the same descriptive-name, no-`(#n)` shape.

## Conventional code review

- [PASS] Diff scope and surrounding call paths — read in full: `adjacent.go` (new, 189 lines), `context.go`, `payload.go`, the three `internal/auth` files, both new test files, the rewritten `context_test.go`, all six scripts, `Makefile`, `.gitignore`, the fragment and the new doc. Call path traced from `resolveMCPRequestHeadersForInvocation` (`internal/app/agent_metadata.go:232`) and `doctor_command.go:41` through `Resolve` → `resolveLibraryPath` → `MaterializeAdjacent`.
- [PASS] Error handling and failure paths — one neutral sentinel error (`adjacent runtime payload unavailable`) on every failure branch, so no path or I/O detail leaks. `writeOwnership`'s unchecked `json.Marshal` is correctly justified inline (the record holds only JSON primitives).
- [PASS] Concurrency, cancellation, cleanup, lifecycle — non-blocking `flock` serializes publishers; `defer held.Close()` and `defer os.RemoveAll(stage)` cover every path. I walked the partial-failure interleavings: a failed rename after the library was already published leaves ownership `pending` with `ps` missing, and the next invocation re-extracts and commits `ready`, matching the injected-failure test. A loaded Windows DLL makes the `old-` rename fail, which correctly degrades to cache rather than corrupting the directory. No stage directories leaked across seven warm runs.
- [PASS] Security, permissions, temporary files, command execution — `root` must be absolute and a non-symlink real directory; `lstat` is used throughout so links are never followed; the pre-publication walk rejects any symlink or non-regular file, including links planted *inside* a previously owned `ps`. The ownership record is decoded with `DisallowUnknownFields`, a trailing-JSON check, an 8192-byte cap, an owner literal, a state enum, `format_version`, target/library agreement, `ps_file_count == 123`, an 8-digit version and 32-byte hex digests. Staged files are `0600`/`0700`, extraction uses `O_CREATE|O_EXCL`, and `inspectAt` bounds `offset+capacity <= len(data)` with `capacity >= containerHeader+size`, so the `container[containerHeader:containerHeader+descriptor.Size]` slice cannot panic. No command execution is introduced.
- [PASS] Backward compatibility and configuration defaults — no flag, command or config surface changed. The removed sidecar discovery is dead code per the repo's own release verification. `PayloadVersion` is part of the cache path, so previous-version caches are simply unused rather than misread.
- [PASS] Platform-specific files and build tags — no build-tag changes; `native_unix.go` / `native_windows.go` / `native_unsupported.go` untouched. All six `CGO_ENABLED=0` targets cross-compile, and CI `Test (Windows)`, `Test (cross-platform compile)` and `Coverage (Windows)` are green.
- [FAIL] Test coverage quality — coverage of *changed statements* is genuinely 100% (V4) and the tests are unusually strong: a real subprocess for the cross-process lock, real symlinks, adversarial token bytes, injected failures at each rename point, neutral-error assertions. The gap is that every `adjacent_test.go` case builds a **1 MiB fixture container** (`BuildContainer(..., 1<<20)`), so the warm-path cost of the real 13.2 MB payload is structurally invisible to the suite. Nothing in the repo measures it, which is how a 9x regression reached review undetected.
- [PASS] Formatting, static analysis, generated files, dependency scope — `gofmt` clean, CI `Lint` green. No `go.mod`/`go.sum` change, so no new dependency: `adjacent.go` reuses the existing `internal/event/lock`. No generated Schema artifacts committed; the tree stayed clean after generation and builds.
- [PASS] Documentation consistency with the code — every falsifiable claim in `docs/runtime-context.md` that I checked is accurate (V11), including the subtle one that only `k9Xm2pQv` is invoked while the other seven exports are integrity-checked only. No rival names in the new doc.
- [PASS] Guard direction — the PR *adds* two gates to `make policy` rather than weakening any. The flag split is deliberate and correct: CI's `Validate Runtime Payload` installs `llvm binutils file` and runs `check-runtime-payload.sh` **without** `--allow-unsupported-tools`, so export inspection is strict in CI and only lenient for local runs. No guarded set shrinks.

## CI and baseline comparison

- [PASS] Required CI checks — all green on the current head `cbc17657`. Taking the **latest attempt per check name** rather than the raw list: 33 `success`, 25 `skipped`, **0 failure** — `Test` (14:10:23Z), `Coverage` (13:59:15Z), `Policy`, `Validate Runtime Payload`, `Interface Integrity`, `Lint`, `Edition`, `CLI Smoke`, `Mock MCP`, `AI Behavior`, all race lanes and all coverage shards. Every `Fail fast (*)` aggregator is `skipped`, which is the "nothing failed" state. For the record: a raw `check-runs` query does show two `failure` rows — the aggregate `Test` and `Coverage` from check suite `92734388781` at 13:48:2x — but both were superseded by successful re-runs, so the current state is green. `Draft Fast Gate` is `skipped`, confirming this is not a draft-only lightweight run.
- [PASS] Claimed known failures — the three plugin/Gitee fixture failures the author describes did not reproduce in either of my two runs (132 `ok` both times); the author already states they were concurrency artifacts that passed on rerun, which is consistent.
- [PASS] Base comparison — the V10 mismatch is **not attributable to this PR**. Classification: environment-dependent / flaky under memory pressure, confirmed by base failing the same check.

  Five runs on the same 18 GiB host:

  | Revision | Command | `internal/app` | `internal/cli` |
  |---|---|---|---|
  | HEAD | `go test -p 2 ./...` (full suite, run 1) | ok 445.537s | **FAIL 16.086s** — `run cmd_schema_catalog: signal: killed` |
  | HEAD | `go test -p 2 ./...` (full suite, run 2) | ok 486.465s | **FAIL 14.736s** — byte-identical message |
  | HEAD | `go test -count=1 ./internal/cli/` (isolated) | — | **ok 135.603s** |
  | HEAD | `go test -p 2 -count=1 ./internal/app/ ./internal/cli/` | ok 352.810s | **ok 126.760s** |
  | BASE | `go test -p 2 -count=1 ./internal/app/ ./internal/cli/` | ok 467.486s | **FAIL 14.199s** — byte-identical message |

  The decisive row is the last one: the **base revision fails the same pairing that head passes**, with the identical `signal: killed` at the same ~14 s mark. A failure that reproduces on base and does not reproduce on head cannot be a PR regression. It is also not deterministic in either direction — head failed it twice under the full suite and passed it both in isolation and under the narrower pairing — so the honest classification is flaky under memory pressure rather than a stable baseline failure.

  Supporting points, all independent:
  1. `internal/cli` and its `schema_testmain_test.go` are **untouched** by this PR.
  2. The failure is `signal: killed` — SIGKILL of the `cmd_schema_catalog` subprocess, not an assertion. The `go build` step succeeded; the *run* step died. That is the signature of memory pressure.
  3. It always dies at ~14–16 s, precisely while the `internal/app` test binary is resident. That binary alone holds **2.12 GiB RSS** (measured directly) on an 18 GiB host that was already paging (2.19 M pageouts, 33 % free) with nothing else running.
  4. The PR did not make the killed process heavier: `cmd_schema_catalog` is 51,715,554 bytes on head vs 51,696,978 on base (+18 KB), and the embedded payload assets keep their fixed capacity (6,291,456 bytes before and after).
  5. `internal/cli`'s `TestMain` compiles *and runs* a second full-app-tree process while `-p 2` keeps `internal/app` alive — two full app trees plus the Go compiler on an 18 GiB host.

  The author's claim is therefore not false, just **unqualified**: it does not record the memory headroom `-p 2` needs on a host of this size. For what it is worth, CI runs these packages as separate sharded jobs, which is why CI is green.

## Findings and blockers

- Blocking code findings: **1** — `internal/runtimepayload/adjacent.go:102-117`. The warm reuse path performs a full extraction plus two full verification passes before it checks whether the already-published resources are reusable.

  Measured on darwin/arm64 against the **real** embedded payload, 7 warm iterations each, same container:

  | Path | avg | min | max |
  |---|---|---|---|
  | `MaterializeAdjacent` (new, warm) | **131.4 ms** | 127.8 ms | 134.5 ms |
  | `Materialize` (pre-existing cache path, warm) | **14.4 ms** | 13.9 ms | 15.8 ms |

  **9.1x**, ~117 ms added per invocation. The tight min/max spread shows it is systematic, not noise.

  End to end, `dws doctor --json` → `detail.elapsed_ms`, five runs each from an isolated install directory:

  | | run 1 | run 2 | run 3 | run 4 | run 5 |
  |---|---|---|---|---|---|
  | HEAD | 308 | 141 | **368** | 352 | 134 |
  | BASE | 246 | 20 | 20 | 21 | 20 |

  Base settles at ~20 ms; head sits at 134–368 ms. The head spread is consistent with the cost of writing 124 fresh files per invocation — on this host an endpoint-security agent scans newly written executables, a plausible source of the 368 ms tail.

  Why it matters: `Resolve()` is reached from `resolveMCPRequestHeadersForInvocation`, i.e. **every MCP business command**, plus both login flows and `dws doctor`. `--dry-run` is correctly exempt. So this adds >100 ms to essentially every real invocation of the CLI.

  Mechanism, in execution order:
  1. `adjacent.go:102` creates a staging directory inside `root`.
  2. `adjacent.go:107` `extractPayload(...)` — **unconditionally** gunzips and untars the whole bundle, writing 124 files (~13.2 MB: a 1.73 MB library plus 11.5 MB across 123 `ps` files) beside the executable.
  3. `adjacent.go:110-113` `readManifest(stage)` + `validateRootManifest(stage, …)` — SHA-256 over all 124 staged files.
  4. `adjacent.go:115` the reuse check finally runs, calling `validateRootManifest(root, …)` — SHA-256 over all 124 published files again.
  5. On reuse, the `defer os.RemoveAll(stage)` at `adjacent.go:106` deletes all 124 staged files.

  So in the steady state steps 2, 3 and 5 are pure waste: ~13.2 MB written, ~26.4 MB hashed, ~13.2 MB deleted, to reach a conclusion the ownership record already implied.

  Suggested direction — the exact shape is your call. `digest` (`adjacent.go:114`) comes from `descriptor.SHA256`, which `Inspect` already validated in memory, and `current.PayloadSHA256` / `current.Manifest` are both available from `adjacent.go:70`. All three exist *before* line 102. One security property must be preserved: comparing `current.Manifest == expected` against the manifest **read from the embedded container** is what stops a crafted on-disk ownership record from validating attacker-chosen library and `ps` bytes, so the trusted manifest still has to come from the container. That does not require materializing 123 data files — either stream just `manifest.json` out of the archive (gunzip plus a tar walk, no writes), or short-circuit on `current.PayloadSHA256 == digest` before staging and fall through to the full extract-and-republish path only when the digest differs or `validateRootManifest(root, current.Manifest, …)` fails. Either keeps the warm path near the base cost (one `validateRootManifest(root, …)` pass, ~14 ms) with identical fail-closed semantics.

  Please add a regression guard alongside the fix, since the existing 1 MiB fixture cannot see this.

- Description blockers (template-required only): none — T1–T8 all pass against `.github/pull_request_template.md`.
- Fragment blockers: none — F1–F6 all pass, 0 PR-introduced CJK or rival-name hits.
- Validation mismatches: **V10**, classified **flaky / environment-dependent** — the base revision fails the same `internal/cli` check under the same pairing that head passes, so it is neither a PR regression nor a false claim, just an unqualified one.
- Not-verified items: **V8** — needs a goreleaser-produced `dist/` plus real npm and Homebrew installs, unavailable in this environment. CI's release-contract and runtime-payload jobs are green on this SHA, and the packaging-side changes here are limited to version strings I verified are complete.
- Non-blocking observations, no action required for this review:
  1. `/api/status` now returns the complete `browserURL` including `callerUmt`, where it previously returned the bare public `AuthorizeURLForLoginRegion(...)` constant. That handler has no `Host`/`Origin` validation — a **pre-existing** property of the loopback server, not something this PR introduces — but a DNS-rebinding attacker would now additionally obtain the runtime token. Impact is low: the value is a device UMID that DWS already sends as the `x-dingtalk-ext` header on every business request, not a credential granting access. If it is cheap, a `Host` allowlist (`127.0.0.1:<port>` / `localhost:<port>`) would close the class of issue rather than this one instance.
  2. `callerUmt` in a browser-launched URL will land in browser history. This looks inherent to attaching context to a browser login URL and therefore intended; noted only so the trade-off is on the record. The DWS-side redaction (terminal, manual links, logs, diagnostics) is correct and well tested.
  3. A SIGKILL between staging and the deferred `RemoveAll` leaks a `.dws-runtime-stage-*` directory (~13 MB) beside the executable, and nothing garbage-collects stale stage directories left by earlier runs. Rare, and `.gitignore` keeps it out of a dev tree, but sweeping orphaned `.dws-runtime-stage-*` while holding the lock would complete the recovery story.
  4. Two dws builds with different payload versions sharing one directory would republish over each other on every invocation. Not a supported layout (Homebrew's `libexec` holds exactly one binary), so noted for completeness only.

## Re-review condition

One item: make the warm reuse path in `MaterializeAdjacent` skip the unconditional extraction and the staged verification pass, so a reuse hit costs roughly one `validateRootManifest(root, …)` as the cache path does today, while keeping the trusted-manifest comparison against the embedded container. Please include a before/after latency measurement over the real embedded payload (not the 1 MiB fixture) and a regression guard so the cost cannot silently return.

Nothing else needs to change — the description, the release fragment, the payload integrity work, the test suite and CI all pass as written, and the ownership / symlink / lock safety design is solid.

Optional and not blocking: qualify the full-suite claim with the memory or parallelism conditions under which `-p 2` is clean, since `internal/cli`'s `cmd_schema_catalog` subprocess is SIGKILLed under concurrency on an 18 GiB host on both this head and its base.

typefield:
@zearlin @haofeng0705 这个 PR 合入后 main 出现一个 CI 回归，所有对着当前 main 的 PR 都会被卡住，麻烦看下。

## 根因

payload 目录从 `20260825` 重命名为 `20260908`（commit 56b14f543），但 `scripts/build/linux-abi/main_test.go:20` 仍硬编码旧版本号：

```go
path := filepath.Join("..", "..", "..", "third_party", "runtimepayload", "20260825", "linux", arch, "libx7k2m9p4q1w8.so")
```

当前 main 树上只有 `third_party/runtimepayload/20260908/linux/{amd64,arm64}/libx7k2m9p4q1w8.so`（文件本身搬运无损，仅路径引用未更新）。

## 影响

所有 PR 的 `Test (race: remaining)` 与 `Coverage (current: remaining)` 两个 job 失败于 `TestCrossPlatformCoverageLinuxABIReleaseLibraries`：

```
bundled library rejected: 1, ../../../third_party/runtimepayload/20260825/linux/amd64/libx7k2m9p4q1w8.so: open ...: no such file or directory
```

已确认的两个受影响案例（均与各自 PR 的改动无关）：
- #1331：https://github.com/DingTalk-Real-AI/dingtalk-workspace-cli/actions/runs/34313114982
- `codex/chat-routing-lifecycle-p0` 分支：完全相同的报错

另外 main 分支自身尚未暴露此问题——2780b9bbf 上最新的 success run 是 Reviewer routing（workflow_run），真正的测试 CI 还没在新 main 上跑过。

## 建议修复

最小改动：`main_test.go:20` 的 `20260825` → `20260908`（一行）。

更稳的做法：测试从 `third_party/runtimepayload/` 扫描实际版本目录（或读 `manifest.json` 的 `payload_version`），避免下次 payload refresh 再次踩中。

说明：`internal/runtimepayload/adjacent_test.go:64` 的 `PayloadVersion = "20260825"` 是刻意构造 version-mismatch 场景的 fixture 值，无需改动。

PR #1296, opened 2026-09-05
perf: compact complete CLI tree; Schema cache for injected-identity tests

typefield:
## Summary

- Keep **one `dws` binary, one command process, and one complete Cobra runtime tree** for every public invocation that passes artifact preflight (root help, version, Schema, utilities, business commands, completion). The sole process-model exception is telemetry: the command process may spawn one detached, short-lived sender process from the same binary via the private `--_dws-telemetry-worker=1` argv, which is handled before Cobra assembly and can never become a business command.
- **Production does not produce or embed Schema identity at compile/release time.** On supported targets production enables the persistent Schema cache with `AllowGenerate: true`: a machine-local `identity.json` is loaded when present, otherwise the first schema-consuming path generates identity from live declarations, publishes protobuf shards, and writes the sidecar. Miss/corruption repairs through the same live assembly path.
- Compact typed metadata and shared builders reduce complete-tree allocations **without** argv-selected product trees, projections, or a second public executor.
- Telemetry does **not** wait on exit: the command process never initializes the telemetry SDK. The completed event (≤4 KB JSON, no argv/stdout/cwd/credentials) is handed off over a bounded 10 ms pipe to a detached, short-lived sender process from the same binary (6 s hard lifetime, ≤8 concurrent send slots); an incomplete handoff is discarded, so the last event is expected to be lost (reliable delivery is the durable outbox follow-up). Business cleanup, signals, and exit codes stay synchronous.
- Self-upgrade invalidates persisted `…/dws/schema/**/identity.json` (and legacy `identity.*.json`) after binary replace so binary B cannot keep serving binary A's Schema. Windows shared cache DACL is writable only by Administrators/SYSTEM (Builtin Users read); creator SID is not granted `GENERIC_ALL`.
- Withdrawn earlier approaches stay withdrawn: launcher/dual-binary, selective product trees, pre-Cobra Schema argv fast path, root-help snapshot/projection, compile-time Schema identity seal, fingerprint-keyed sidecars.

Normative notes: [`docs/rfc-schema-runtime-cache.md`](docs/rfc-schema-runtime-cache.md), [`.changes/schema-runtime-cache.md`](.changes/schema-runtime-cache.md).

## Architecture decisions

- Public process argv does **not** select a product factory or utility-only tree; the private telemetry sender invocation is the only pre-Cobra argv path and cannot become a business command.
- Root help renders the final runtime tree (no sealed help snapshot).
[...]

PR #1364, opened 2026-09-11
fix(install): keep shared-cache re-share and readability checks space-safe

typefield:
## Summary

- `for identity_file in $(find ...)` word-splits paths, so a custom `DWS_SCHEMA_CACHE_SHARED_DIR` containing spaces (e.g. "…/My Shared Cache") broke the shared-cache warm-up in two places:
  - the fresh-edition re-share `chmod` loop ran on split bogus paths and failed, falling back with a spurious "not shared" warning;
  - `shared_schema_artifacts_readable` iterated the same way, so every check passed vacuously on garbage paths — the readability guard was hollowed out and could advertise a shared cache other users cannot read.
- Both sites now pass sidecar paths as `find -exec sh -c … sh {} +` arguments, which never word-split. POSIX-sh safe (verified under `dash` semantics via the existing `sh` harness tests).

**Stacked on #1296** — the touched code (re-share loop, readability guard) only exists on that branch. Until #1296 merges this PR's diff will show the whole stack; afterwards it collapses to exactly this fix. Merge only after #1296.

## Verification

- [x] `bash -n scripts/install.sh`
- [x] New regression tests: spaced custom root end-to-end (success message, no "not shared" fallback, edition o+rx, shards o+r) and the guard firing on an unreadable 0600 artifact under spaced paths (`READABLE_FAIL`) while a readable spaced tree passes (`READABLE_OK`).
- [x] Regression-proof: with only `scripts/install.sh` reverted, the new tests fail; with the fix they pass.
- [x] Full `TestInstallScript*` suite green (14.8s).

## Risk tier

- [ ] **High-risk**
- [x] Low-risk: POSIX installer hardening; default shared paths (`/var/cache/dws`, `/Library/Caches/dws`, `%ProgramData%\dws`) contain no spaces and are unaffected; failure direction stays fail-closed (artifacts remain private, nothing extra is exposed).

## Test plan

- [x] `go test ./test/scripts/ -run TestInstallScript`
- [ ] CI green (required checks will run once unstacked after #1296 merges).


typefield:
取消叠加方案：新复审（5177179733）已将空格路径与父目录可达性列为 #1296 的阻塞项，修复将直接进入 #1296。本 PR 关闭，修复内容以 #1296 上的提交为准。

